PR #9, opened 2013-04-03
Performance improved with new _iter_category and _iter_categories method...

boudewijn-tribler:
Code cleanup improves performance of _iter_category and _iter_categories.  Testcases still pass.

### DIFF
--- a/community.py
+++ b/community.py
@@ -1295,65 +1295,63 @@ class Community(object):
 
     def _iter_category(self, category):
         while True:
-            no_result = True
-            
+            index = 0
             keys = self._candidates.keys()
-            key_index = 0
-            key_value = None
 
-            while True:
-                try:
-                    key_value = keys[key_index]
-                    candidate = self._candidates[key_value]
-                    
-                    if candidate.in_community(self, time()) and candidate.is_any_active(time()) and category == candidate.get_category(self, time()):
-                        no_result = False
-                        yield candidate
-                        
-                        keys = self._candidates.keys()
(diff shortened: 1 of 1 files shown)


boudewijn-tribler:
Patch included to only yield None when none of the candidates match.

As for the `index -= 1` logic, candidates are added in sequence (since `_candidates` is an `OrderedDict`).  And we want to iterate over these candidates in order.  When a candidate _that we have not yet iterated over_ (`if keys[index] != key` is `True`) is removed we assume that only a single candidate was removed and set the index back by one to ensure the following `index += 1` does not result in us skipping a candidate.

Starting from the beginning of the candidates would result in the earlier candidates being yielded more frequently whenever a candidate is removed.  I believe this is the opposite of what `_iter_category` is supposed to achieve.

Finally, ValueError is raised by `keys.index(key)` when `keys` does not contain `key`.  In other words, the candidate associated with `key` was removed.

NielsZeilemaker:
Ok great, the fix for yielding None is fine. 
The index -=1 stuff, I guess it makes sense, however I don't know if the assumption of a single candidate being removed always holds.

boudewijn-tribler:
Indeed, that assumption will not always hold.  However, I believe it is better than starting at the beginning (which will always result in a bias towards the earlier candidates) or only doing the `index += 1` (which will always skip a candidate on this iteration).

Of course we could compare the old `keys` with the new `keys` and exactly figure out what happened.  But this seems overkill to me.

PR #36, opened 2013-04-24
Code cleanup

boudewijn-tribler:
API change: https://github.com/boudewijn-tribler/dispersy/commit/e79990417bb3555aa2816a4ed79d8c5c8fdcfe10

Code cleanup: https://github.com/boudewijn-tribler/dispersy/commit/8d83f5e5bb647d7d3195b72891f999fa086e2a9e

Replaced DISPERSY_DEBUG_DATABASE_QUERIES enviromental variable with
--explain-query-plan and --runtime-statistics.

--explain-query-plan will enable the attach_query_plan decorator.
This decorator will run EXPLAIN QUERY PLAN once for every unique query
and output the results to a logger called "explain-query-plan".

--runtime-statistics will enable the attach_runtime_statistics
decorator.  This decorator will count how often the function is called
and how much wall-time was used.  A logger called "runtime-statistics"
will get the results when the python application exits.

Attach_runtime_statistics tracks statistics per string, not per
function (what a profiler does).  The string can be modified based on
the arguments of the function, i.e. for the database we track
different statistics per query by adding the sql statement in the
string.  See https://github.com/boudewijn-tribler/dispersy/commit/e509cb1da2d3925cb636d2edd897f5f1780023bb for more Info.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@
 *.pyo
 *~
 \#*\#
-
+*_flymake.py
--- a/database.py
+++ b/database.py
@@ -9,20 +9,34 @@ This module provides basic database functionalty and simple version control.
 import logging
 logger = logging.getLogger(__name__)
 
-from os import environ
 from sqlite3 import Connection, Error
+
+from .decorator import attach_runtime_statistics
 
 if __debug__:
     import thread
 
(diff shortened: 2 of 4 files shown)


boudewijn-tribler:
Feedback please

Issue #43, opened 2013-04-24
Dispersy.stop() is unloading communities in random order.


boudewijn-tribler:
We should remember the order that communities are defined using `Dispersy.define_auto_load()` and use its reverse order when unloading communities when ``Dispersy.stop()` is called.

This will prevent bugs from occurring when communities depend on each other, i.e. in a future database cleanup we want AllChannelCommunity to create a database that all ChannelCommunities will need.

PR #47, opened 2013-04-26
Unload communities in reversed auto load order.

boudewijn-tribler:
https://github.com/Tribler/dispersy/issues/43


boudewijn-tribler:
Bump

PR #53, opened 2013-05-08
Fixes exception.

boudewijn-tribler:
An undocumented IndexError can occur in low level inspection code, if
it does, we now assume the source file and line numer are unknown.


boudewijn-tribler:
Accepting this change.  The fewer pull requests we have open will reduce the spam tribler-ci is causing :p

Issue #56, opened 2013-05-16
Infinite loop when yielding introduction candidates.


boudewijn-tribler:
An infinite loop bug was introduced in: https://github.com/Tribler/dispersy/commit/11c85c9b6e3fc710cbfaa70c63a68fd1e7c0ab46#community.py

Caused by:

``` python
if (result == exclude_candidate) or (exclude_tunnel and result.tunnel):
    continue
```

I am extending test_candidates.py to reproduce before fixing.

boudewijn-tribler:
A related bug: result can be None, hence the line should have been

``` python
if (result == exclude_candidate) or (exclude_tunnel and result and result.tunnel):
    continue
```

boudewijn-tribler:
Fixed

PR #62, opened 2013-06-03
Changes candidate iterator behavior and adds unit tests. 

boudewijn-tribler:
All dispersy_yield_..._candidates methods now behave the same way:
- take no arguments
- yields only Candidate instances
- stops when no more Candidates are available

Having the above behavior makes it much easier to treat them as
Candidate iterators.  Normal python tools such as cycle and ifilter
can be used to obtain more candidates and exclude certain candidates
when needed.


boudewijn-tribler:
Fixes: #56.

Tested using Project http://jenkins.tribler.org/jenkins/job/Experiment_AllChannel+Channelcommunity_devel_boudewijn/

Issue #7, opened 2013-03-26
RequestCache set/has


NielsZeilemaker:
I got a community which uses the RequestCache to store requests. When receiving a message, I check if the RequestCache has any requests with this identifier stored, if not i accept it.

However, if the RequestCache does not have this type of Cache stored but a different one. The has-method will return false, but the set-method will not allow me to store the new Cache.

Can we improve upon this? Why is the request-cache global?

boudewijn-tribler:
Cashes should be identified by ID and type (currently it only uses ID, causing the problem you describe above). We may event want to give each type its own request cache instance.  That would simplify the code.

There is no longer a reason for the request cache to be global (historically Dispersy handled all messages, we are slowly moving away from this, i.e. making Community responsible).

NielsZeilemaker:
Seems like a nice solution. Currently I can't run my experiments in debug mode due to this, and in optimized mode this is causing stuffs as the clashes are not checked.

boudewijn-tribler:
Does everyone still agree with the proposed solution?
- remove the type stuff in the cache
- have one cache for each type
- have each cache as a Community member

NielsZeilemaker:
The only problem I can think of is inheritance. Currently both the has and get method allow you to pass a parent to find a child class. If we create separate caches for each type, this won't work anymore.

boudewijn-tribler:
Using either has or get to obtain the cache associated with some ID would result with whatever cache is stored.  Things should be fine as long as the stored cache is compatible with what the code expects, i.e. has the required properties and methods.

Your point is still valid, but do we have a better option?

NielsZeilemaker:
Could we not have a single requestcache, which has multiple cache for types. But now types arn't defined by the class you pass, but by an unicode identifier very much similar to all other identifiers?

Seems like a cleaner fix to me, and won't require much refactoring. As we still have a single request_cache member in each community. As opposed to 10, a single requestcache for each type of requests.

boudewijn-tribler:
Very true.  The only downside that I can think of having a single requestcache is more code and more parameters for its API.  With 10 'single requestcaches' we won't need to pass and store the string that indicates the type, since it is one type per cache.

Effectively, do we call

``` python
# many small caches
self.request_cache_A.get(random_number)
self.request_cache_B.get(random_number)
```

or

``` python
# single big cache
self.request_cache.get(random_number, "A")
self.request_cache.get(random_number, "B")
```

[...]

Issue #3, opened 2013-03-18
Timeout-problem in RequestCache


NielsZeilemaker:
During the DAS4 runs I often get this error: 
Looking at the code, this can only be caused by calling the _on_timeout method twice or if the _on_timeout method is still called after the unregistering it.

``` python
"callback.py", line 613, in _loop
    result = call[0](*call[1], **call[2])
"requestcache.py", line 103, in _on_timeout
    cache = self._identifiers[identifier]
KeyError: 33437
```

whirm:
@boudewijn-tribler could you take a look at this one please? its ben sitting there for 3 months now. thx

boudewijn-tribler:
This issue will be difficult to track down, I suggest fixing #7, I suspect that this will either solve this problem or at the very least make this problem much easier to diagnose.

PR #66, opened 2013-06-12
RequestCache bug unittest

NielsZeilemaker:
Unittest which is triggering the bug as reported in Tribler/dispersy#3 and Tribler/dispersy#7

### DIFF
--- a/tests/test_requestcache.py
+++ b/tests/test_requestcache.py
@@ -27,4 +27,5 @@ class TestRequestCache(unittest.TestCase):
         if not self.request_cache.has(identifier, Cache2):
             self.request_cache.set(identifier, Cache2())
 
-        assert self.request_cache.get(identifier, Cache1), 'should not have replaced the first assigned cache'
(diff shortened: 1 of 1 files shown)


boudewijn-tribler:
If we add the `unittest.expectFailure` decorator we can merge this pull.  We can remove the decorator once it is fixed.

http://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

NielsZeilemaker:
Doesn't make sense to me. It's a bug which is causing problems during my experiments for over 4 months already. This needs to be fixed asap.

whirm:
@boudewijn-tribler, @NielsZeilemaker: Any news on this?

whirm:
According to @boudewijn-tribler  #106 fixes this among other things. (But also changes the API and breaks Tribler and I guess @NielsZeilemaker scenarios.)

@NielsZeilemaker Do you agree #106 supersedes this pull request?

NielsZeilemaker:
Be my guest, I don't mind what the solution will be. As long as it gets fixed.

PR #76, opened 2013-06-28
Fixes and a live overlay test.

boudewijn-tribler:
Fix to callback: https://github.com/boudewijn-tribler/dispersy/commit/5282b1483ec44ae6e5d22708192243326f187c13

Fix to Dispersy yield behaviour: https://github.com/boudewijn-tribler/dispersy/commit/eddf9f0cf2723e17f3e09772872b3ae601b4be10

Fix for tracker bug: https://github.com/boudewijn-tribler/dispersy/commit/4325692fe7865518e8a4690c2ea6efccd10d77a3


boudewijn-tribler:
retest this please

dispersy.py
```diff
@@ -2418,6 +2417,8 @@ def create_introduction_request(self, community, destination, allow_sync, forwar
                 self._statistics.walk_bootstrap_attempt += 1
             if request.payload.advice:
                 self._statistics.walk_advice_outgoing_request += 1
+            if self._statistics.are_debug_statistics_enabled():
```
NielsZeilemaker: Maybe use the dict_inc method as used in other parts of the code.
This will check if statistics are enabled + increment a value.

statistics.py
```diff
@@ -97,6 +126,16 @@ def enable_debug_statistics(self, enable):
                 self.bootstrap_candidates = {}
                 self.overlapping_stumble_candidates = {}
                 self.overlapping_intro_candidates = {}
+
+                # SOURCE:INTRODUCED:COUNT nested dictionary
+                self.received_introductions = defaultdict(lambda: defaultdict(int))
```
NielsZeilemaker, line 128: I like the defaultdicts, but maybe switch all other dict to this as well?

tests/debugcommunity/conversion.py
```diff
@@ -9,19 +9,23 @@ class DebugCommunityConversion(BinaryConversion):
...
-        self.define_meta_message(chr(8), community.get_meta_message(u"full-sync-text"), self._encode_text, self._decode_text)
-        self.define_meta_message(chr(9), community.get_meta_message(u"ASC-text"), self._encode_text, self._decode_text)
-        self.define_meta_message(chr(10), community.get_meta_message(u"DESC-text"), self._encode_text, self._decode_text)
-        self.define_meta_message(chr(11), community.get_meta_message(u"last-1-doublemember-text"), self._encode_text, self._decode_text)
-        self.define_meta_message(chr(12), community.get_meta_message(u"protected-full-sync-text"), self._encode_text, self._decode_text)
-        self.define_meta_message(chr(13), community.get_meta_message(u"dynamic-resolution-text"), self._encode_text, self._decode_text)
-        self.define_meta_message(chr(14), community.get_meta_message(u"sequence-text"), self._encode_text, self._decode_text)
-        self.define_meta_message(chr(15), community.get_meta_message(u"full-sync-global-time-pruning-text"), self._encode_text, self._decode_text)
+    def __init__(self, community, version="\x02"):
```
NielsZeilemaker, line 12: Why is the debugcommunity at version 2.0 actually?

tests/test_overlay.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @skipUnless(environ.get("TEST_OVERLAY") == "yes", "This 'unittest' tests the health of a live overlay, as such, this is not part of the code review process")
+    def test_barter_community(self):
+        return self.check_live_overlay(cid_hex="4fe1172862c649485c25b3d446337a35f389a2a2",
+                                       version="\x01",
+                                       enable_fast_walker=False)
+
+    @skipUnless(environ.get("TEST_OVERLAY") == "yes", "This 'unittest' tests the health of a live overlay, as such, this is not part of the code review process")
+    def test_search_community(self):
+        # TODO because the search community uses a modified search message, the introduction request is larger than
```
NielsZeilemaker: This actually not true, the searchcommunity replaces the bloomfilter with a bloomfilter which contains preferences.
This replacement is actually smaller than the original one.

statistics.py
boudewijn-tribler, line 128: I agree, would prefer to have all dicts using defaultdicts.

dispersy.py
boudewijn-tribler: If we use defaultdicts everywhere, checking are_debug_statistics_enabled seems a logical replacement for the check in dict_inc.

As for why I am using this check here, I wrote it to be similar to the python code a few lines lower where it checks if a statistics parameter is None, using the actual 'are...enabled' check seemed leaner though.

tests/debugcommunity/conversion.py
boudewijn-tribler, line 12: I probably increased it when one of the existing messages changed format.  But since all unit tests are run on an in-memory database this is no longer needed.

dispersy.py
NielsZeilemaker: The nice thing of comparing it to None in the dict_inc, is that it allows you to enbable a single statistic vs only being able to enable all.

tests/test_overlay.py
boudewijn-tribler:
``` python
raise DropPacket("Invalid packet size (there are unconverted bytes)")
```

the above raise occurred, so the signature was not where it should be in a standard introduction-request.  This usually indicates that there were bytes left in the packet after the conversion, though perhaps there is something else going on this this case.

According to community/search/conversion.py there are a few bytes added `pack('!IBH',...`.  These seem to be the additional bytes.

NielsZeilemaker: I add bytes after disabling the actual bloomfilter, hence the overall packet size is less.

dispersy.py
boudewijn-tribler: I've changed the check to check for None.

dispersy.py
```diff
@@ -2418,6 +2417,8 @@ def create_introduction_request(self, community, destination, allow_sync, forwar
                 self._statistics.walk_bootstrap_attempt += 1
             if request.payload.advice:
                 self._statistics.walk_advice_outgoing_request += 1
+            if self._statistics.outgoing_introduction_request != None:
```
NielsZeilemaker: Please change it to self._statistics.dict_inc(self._statistics.outgoing_introduction_request, destination.sock_addr) for consistency purposes. The dict_inc is used everywhere.

tests/test_overlay.py
boudewijn-tribler:
``` python
if placeholder.offset != placeholder.first_signature_offset:
    logger.warning("invalid packet size for %s data:%d; offset:%d", placeholder.meta.name, placeholder.first_signature_offset, placeholder.offset)
    raise DropPacket("Invalid packet size (there are unconverted bytes)")                                                                       
```

The signature is still not where it should be.  The standard conversion.py doesn't try to read the sync bloomfilter (because, as you say, it is disabled).  Hence the additional bytes from `pack('!IBH',...` will never be converted, resulting in the above drop.

NielsZeilemaker: Yup, thats completely true. I would change this comment to:
#TODO because the search community modifies the introduction-request message (it replaces the normal bloomfilter with a bloomfilter with preferences), the incomming requests are dropped.

dispersy.py
boudewijn-tribler: outgoing_introduction_request uses a dict in a dict, this is not compatible with dict_inc.  I'm reluctant to add a dict_dict_inc method.

I do agree with things being consistent.  Considering that we both prefer defaultdicts, the dict_inc may not be needed anymore.

tests/test_overlay.py
boudewijn-tribler: Changed, additionally I added 'while keeping the 'sync enabled' bit at zero' for completeness.

dispersy.py
NielsZeilemaker: That's not true, outgoing_introduction_request is a defaultdict, hence this would still work.

If you change it to:
self._statistics.dict_inc(self._statistics.outgoing_introduction_request, destination.sock_addr)

The dict_inc internally does a:
self._statistics.outgoing_introduction_request[destination.sock_addr] += 1

Which is the same as you wrote on line 2421.

boudewijn-tribler: Ah yes, I was confused with received_introductions.

I'll change the other two (outgoing_introduction_request and incoming_introduction_response) to use dict_inc.

boudewijn-tribler:
Any other comments on this pull request?

NielsZeilemaker:
Only the remaining open remarks:
- Change all dicts to defaultdicts in statistics
- Set the DebugCommunity version to 1 instead of 2

boudewijn-tribler:
I would rather do the defaultsdicts cleanup with another pull as it is unrelated.  I can make an issue for this if you want.

As for the DebugCommunity version 1 v.s. 2, there is no need to change this.

NielsZeilemaker:
The DebugCommunity version 1 vs 2 did cause the problems with the trackers.
The defaultdicts stuff, either stick to the normal dicts, or convert all dicts to defaultdicts. Committing a version which mixes the two does not make sense.

synctext:
Hi Guys, I'll ask Elric to see what needs to be done with this pull request.
Then he can decide on what is needed to finish this request.
So: 1 vs. 2; defaultdicts now or later.

whirm:
I agree with @NielsZeilemaker about defaultdict. 

About the v2 for the debug community, are we positively sure it won't bring any more trouble?

boudewijn-tribler:
Currently we know v2 doesn't cause any problems anymore.  If the version number cause problems elsewhere I'd rather find out running the unit tests than when one of the live communities migrate.

whirm:
If you are sure that we aren't going to cause more trouble to our users, sure, go ahead.

boudewijn-tribler:
As sure as anyone can be.

boudewijn-tribler:
retest this please

tests/test_overlay.py
```diff
@@ -0,0 +1,186 @@
...
+        summary.info(community.cid.encode("HEX"))
+
+        history = []
+        begin = time()
+        for _ in xrange(60 * 15):
+            yield 1.0
+            now = time()
+            info = Info()
+            info.diff = now - begin
+            info.candidates = [(candidate, candidate.get_category(community, now)) for candidate in community.dispersy_yield_candidates()]
```
NielsZeilemaker: Community.dispersy_yield_candidates() will never return candidates which have none as a category. However, you attempt to write candidates which are in the "none" category to file. Is this intentional?

boudewijn-tribler: At some point during debugging I wanted to see if 'none' Candidates were properly cleaned up.  Never noticed that it no longer worked.

I'm thinking about adding a candidates property to Community to make the raw _candidates dict available for this.  Any objections?

NielsZeilemaker: I wouldn't add raw as the candidates with a none category are actually
not-connectable and hence should not be used.

For this test I would simply use _candidates.itervalues().

boudewijn-tribler: Done.  The Jenkins task is running, so we will have non-zero numbers for the none category in a few hours.

PR #83, opened 2013-07-09
Removed GlobalCandidateCache.

boudewijn-tribler:
Since we want all communities to manage their own Candidate, the
GlobalCandidateCache is no longer needed.

Several Candidate related methods have been moved from dispersy.py to
community.py, finally, all references to the GlobalCandidateCache have
been rewritten to use candidates from the community directly.

### DIFF
--- a/dispersy.py
+++ b/dispersy.py
@@ -235,51 +235,6 @@ class MissingSequenceCache(MissingSomethingCache):
         return "-missing-sequence-%s-%s-%s-%d-" % (message.community.cid, message.authentication.member.mid, message.name.encode("UTF-8"), message.distribution.sequence_number)
 
 
-class GlobalCandidateCache():
-
-    def __init__(self, dispersy):
-        self._dispersy = dispersy
-
-    def __contains__(self, item):
-        for community in self._dispersy._communities.itervalues():
-            if item in community._candidates:
-                return True
-
-    def __setitem__(self, key, item):
-        now = time()
-        for community in self._dispersy._communities.itervalues():
-            if item.in_community(community, now):
-                community._candidates[key] = item
-
-    def __delitem__(self, item):
-        for community in self._dispersy._communities.itervalues():
-            if item in community._candidates:
-                del community._candidates[item]
-
-    def iteritems(self):
-        for community in self._dispersy._communities.itervalues():
-            for key, value in community._candidates.iteritems():
-                yield key, value
-
-    def itervalues(self):
-        for community in self._dispersy._communities.itervalues():
-            for value in community._candidates.itervalues():
-                yield value
-
-    def get(self, item, default=None):
-        for community in self._dispersy._communities.itervalues():
-            if item in community._candidates:
-                return community._candidates[item]
-
-        return default
-
-    def __len__(self):
-        candidates = set()
-        for community in self._dispersy._communities.itervalues():
-            candidates.update(community._candidates.itervalues())
-        return len(candidates)
-
-
 class Dispersy(object):
 
     """
@@ -331,11 +286,6 @@ class Dispersy(object):
                 os.makedirs(database_directory)
             database_filename = os.path.join(database_directory, database_filename)
         self._database = DispersyDatabase(database_filename)
-
-        # peer selection candidates.  address:Candidate pairs (where
-        # address is obtained from socket.recv_from)
-        self._candidates = GlobalCandidateCache(self)
-        self._callback.register(self._periodically_cleanup_candidates)
 
         # assigns temporary cache objects to unique identifiers
         self._request_cache = RequestCache(self._callback)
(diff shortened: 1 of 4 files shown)


dispersy.py
```diff
@@ -2909,14 +2766,14 @@ def _forward(self, messages):
...
-                                            for candidate
-                                            in self._candidates.itervalues()
-                                            if any(candidate.is_associated(message.community, member)
-                                                   for member
-                                                   in message.destination.members)],
-                                           [message])
-                       for message
-                       in messages)
+                                     for candidate
+                                     in message.community.candidates.itervalues()
```
NielsZeilemaker, line 2770: I guess this piece of code is translating a member destination to a list of candidates. But it seems a bit dodgy to me. The message.community.candidates dict contains candidates which are scheduled for cleanup, and hence should not be considered active.
Moreover, what are the benefits of using members instead of candidates?

boudewijn-tribler, line 2770: I rewrote this piece of code to behave as it was before, including the admittedly dodgy code.  Iterating over all candidates (including the ones scheduled for cleanup) should not be a problem because messages with MemberDestination should only be sent on-demand, i.e. you know its online.

But I do agree that it would be better to handle this using CandidateDestination instead.  Searching for MemberDestination shows that it is only used for the 'dispersy-signature-request' message, I will make an Issue suggesting the change.

NielsZeilemaker, line 2770: Ok great, it's a bit scary now as you need to keep the candidates you're going to send a signature request message to active without screwing with the walker. That seems like an impossible task.

Maybe we can add a return false if no candidates were found for any of these members. As this could be a silent fail, which is occurring in the bartercast community now.

NielsZeilemaker, line 2770: Sorry about that last comment, it is doing that already. Altough the create_signature_request method is not doing anything with this returnvalue.

boudewijn-tribler, line 2770: False is already returned when no candidates can be found.  Although this may not immediately be obvious from the code.

The bartercast community doesn't check this though, instead, the request will timeout after a few seconds.  So its not an entirely silent fail.  Still, it can definitely use improvement.  See #84.

[edit: yea, in that case my answer doesn't apply anymore either]

NielsZeilemaker, line 2770: I can find the timeout, however I cant see a warning being generated, only a call to remove_from_slope.
Are you sure this is logged somewhere?

boudewijn-tribler, line 2770: Currently no warning is generated, neither is the failure logged.  The main purpose of the bartercast community is to test its new features: byte counting and pruning.  The old statistics that included record creation success rate, among other things, no longer exist because Johan didn't want to actively push to a central server.

NielsZeilemaker, line 2770: You're not worried that after 30 minutes every attempt to create a new barter record fails due to the candidate now not being active anymore, and hence not in the _candidates list?

boudewijn-tribler, line 2770: Not worried at all.  Barter records are made with peers that are 'on the slope'.  And a separate ping/pong message is used to ensure that peers 'on the slope' are actually there.

NielsZeilemaker, line 2770: Yes, I know they are connectable, but are the still in the candidates list? I.e. are they still in the walked, stumble, or introduced category? If so, how do you keep them there?

boudewijn-tribler, line 2770: Now I understand what you mean, and you are absolutely right.  From the top of my head, the best way to fix this is to add candidate.walk_response() in bartercast3/community.py:on_pong.  That way the candidate will stay in the 'walk' category with every successful ping/pong.

Are you doing something similar with the ping/pong in the private search community?

NielsZeilemaker, line 2770: You can't as this will screw with the walker for sure. 
If you want a dodgy fix for this I would suggest to first call get_category and if that returns none call inactive to reset the timestamps to just below a category but above the periodic cleanup timeout.

I'm using candidatedestinations in my community, hence don't have a similar problem.

boudewijn-tribler, line 2770: Using CandidateDestination would be the best solution.  But I'll need to change the Dispersy core for that, as the signature-request is from the core.  Still... better fix this before we do a release.

Can we accept this pull request?  I'll fix the batercommunity separately from the candidate cleanup.

NielsZeilemaker, line 2770: Sure, how about the effort stuff you did for Adele. Did this include the same bug?

boudewijn-tribler, line 2770: Yea, I was thinking the same thing... 

Luckily all the DAS emulations used a very short cycle size, so the candidates would usually still exist (i.e. not yet been removed from _candidates).  Plus, we did keep an eye on the record failure, and it wasn't high, even with churn.

I'm more worried about the live experiment where we had 30 minute cycles, hence this bug must have been more noticeable.  I'll have to check the logs to see the record success rate there.  Though its more difficult to access, as we can't distinguish between failure because of churn or this bug.

Regardless, I'll send her an email.

PR #85, opened 2013-07-10
Candidate instance is no longer aware of communities.

boudewijn-tribler:
previously the candidate instance contained information on a per
community basis.  This meant that most methods contained an argument
to identify the community for which an event occurred.

We have changed the candidate to no longer be aware of communities.
This means that:
- all arguments specifying a community have been removed,
- some methods have been removed entirely (i.e. those that exclusively
  worked on all communities), and
- some methods have been rewritten to properties (i.e. those that no
  longer have any arguments).

All changes outside candidate.py reflect changes to handle these API
changes.

### DIFF
--- a/candidate.py
+++ b/candidate.py
@@ -57,17 +57,6 @@ class Candidate(object):
         assert is_address(wan_address), wan_address
         return self._sock_addr
 
-    def get_members(self):
-        # TODO this method should be removed, however, it is possibly still used from allchannel/community.py and privatesearch/community.py
-        logger.warning("DEPRECATED.  Please call community.get_candidate(sock_address).get_members() instead!")
-
-        # preferably use the WalkerCandidate directly
-        candidate = community.get_candidate(self._sock_addr)
-        if candidate:
-            return candidate.get_members(community)
-        else:
-            return []
-
     def __str__(self):
         return "{%s:%d}" % self._sock_addr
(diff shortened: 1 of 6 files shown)


boudewijn-tribler:
Once we agree on the details of this pull, I will make a new pull request with the required changes to Tribler.

candidate.py
```diff
@@ -57,9 +57,12 @@ def get_destination_address(self, wan_address):
         assert is_address(wan_address), wan_address
         return self._sock_addr
 
-    def get_members(self, community):
+    def get_members(self):
+        # TODO this method should be removed, however, it is possibly still used from allchannel/community.py and privatesearch/community.py
```
NielsZeilemaker: Both the allchannel and privatesearh is using the "other" get_members method.

boudewijn-tribler: We will need to rewrite those to get the WalkCandidate instance if they only have a Candidate instance.  I'll change that on the Tribler pull.

NielsZeilemaker: They only get candidates from the walker, hence I would expect them to only have walkcandidate instances.

boudewijn-tribler: Ahh ok, so the "other" get_members is the one from WalkCandidate.  In that case I'll remove `def get_members` from Candidate.

PR #87, opened 2013-07-10
Removes MemberDestination.

boudewijn-tribler:
Using MemberDestination can result in messages not being delivered
when the walker did not recently visit the member in question.  Using
CandidateDestination is preferred.

Currently only the signature request uses MemberDestination, we have
rewritten it to use CandidateDestination instead.  Since neither of
these destinations add anything to the binary packet, this change is
fully backwards compatible.

### DIFF
--- a/tests/debugcommunity/community.py
+++ b/tests/debugcommunity/community.py
@@ -5,7 +5,7 @@ from ...authentication import DoubleMemberAuthentication, MemberAuthentication
 from ...candidate import Candidate
 from ...community import Community, HardKilledCommunity
 from ...conversion import DefaultConversion
-from ...destination import CommunityDestination, CandidateDestination
+from ...destination import CommunityDestination
 from ...distribution import DirectDistribution, FullSyncDistribution, LastSyncDistribution, GlobalTimePruning
 from ...message import Message, DelayMessageByProof
 from ...resolution import PublicResolution, LinearResolution, DynamicResolution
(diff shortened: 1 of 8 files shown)


tests/debugcommunity/community.py
```diff
@@ -94,13 +94,16 @@ def create_full_sync_global_time_pruning_text(self, text, store=True, update=Tru
     # double-signed-text
     #
 
-    def create_double_signed_text(self, text, member, response_func, response_args=(), timeout=10.0, forward=True):
+    def create_double_signed_text(self, text, candidates, member, response_func, response_args=(), timeout=10.0, forward=True):
+        assert isinstance(candidates, list)
+        assert len(candidates) == 1
```
NielsZeilemaker: That's a bit strange, why call it candidates and then require the list to be of length 1

tests/debugcommunity/node.py
```diff
@@ -504,16 +504,17 @@ def create_dispersy_missing_sequence(self, missing_member, missing_message, miss
                          destination=(destination_candidate,),
                          payload=(missing_member, missing_message, missing_sequence_low, missing_sequence_high))
 
-    def create_dispersy_signature_request(self, message, global_time, destination_member):
+    def create_dispersy_signature_request(self, message, global_time, destination_member, destination_candidate):
```
NielsZeilemaker: The member is not used, why not remove it?

boudewijn-tribler: The member should be used, i.e. it should be given as an argument to the Authentication policy.

I checked, and there is no unit test where the Node initiates a signature request, hence this code is never called.  I'll see if I can either add a quick unit test, otherwise I'll just remove this unused method.

tests/debugcommunity/community.py
boudewijn-tribler: You are both right.  I forgot this was all rewritten to be double signed only, previously it could create messages with signatures from one or more peers, hence the list.

tests/debugcommunity/node.py
```diff
@@ -504,16 +504,17 @@ def create_dispersy_missing_sequence(self, missing_member, missing_message, miss
...
-        isinstance(destination_member, Member)
+        assert isinstance(message, Message.Implementation)
+        assert isinstance(global_time, (int, long))
+        assert isinstance(destination_member, Member)
+        assert isinstance(destination_candidate, Candidate)
         meta = self._community.get_meta_message(u"dispersy-signature-request")
         return meta.impl(distribution=(global_time,),
-                         destination=(destination_member,),
+                         destination=(destination_candidate,),
                          payload=(message,))
```
NielsZeilemaker, line 518: I can't see the member being passed as an authentication= argument.

tests/debugcommunity/node.py
boudewijn-tribler: Nop, confused again.  The destination_member is an argument of the authentication policy for the double signed message, not the signature request.

In that case the destination_member is indeed no longer needed.

PR #95, opened 2013-07-18
Fixes missing _candidates.

boudewijn-tribler:
Fixes tracker statistics reporting, would previously crash the
periodic report when attempting to use non existing _candidates
property.

Note that we need to reset our trackers and update the statistics parser!

### DIFF
--- a/tool/tracker.py
+++ b/tool/tracker.py
@@ -332,7 +332,7 @@ class TrackerDispersy(Dispersy):
 
             print "BANDWIDTH", self._endpoint.total_up, self._endpoint.total_down
             print "COMMUNITY", mapping[TrackerCommunity], mapping[TrackerHardKilledCommunity]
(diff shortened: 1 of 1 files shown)


tool/tracker.py
```diff
@@ -331,7 +332,7 @@ def _report_statistics(self):
 
             print "BANDWIDTH", self._endpoint.total_up, self._endpoint.total_down
             print "COMMUNITY", mapping[TrackerCommunity], mapping[TrackerHardKilledCommunity]
-            print "CANDIDATE", len(self._candidates)
+            print "CANDIDATE2", sum(len(len(list(community.dispersy_yield_verified_candidates())) for community in self._communities.itervalues()))
```
NielsZeilemaker: Minus one len command I think?

boudewijn-tribler: Very true.

boudewijn-tribler:
Some builder failed?  Can you look at this Elric?

boudewijn-tribler:
retest this please

PR #98, opened 2013-07-18
Improvements to conversion selection.

boudewijn-tribler:


community.py
```diff
@@ -1279,51 +1280,64 @@ def get_members_from_id(self, mid):
...
+        Returns the default conversion or one associated with USING.
+
+        When USING is a *string* this method returns the first available conversion that can *decode* USING, this is
+        tested in reversed order using conversion.can_decode_message(USING).  Typically a conversion can decode a string
+        when it matches: the community version, the Dispersy version, and the community identifier, and the conversion
+        knows how to decode messages types described in USING.
+
+        When USING is a *Message* or *Message.Implementation* instance this method returns the first available
+        conversion that can *encode* USING, this is tested in reversed order using conversion.can_encode_message(USING).
+        Typically a conversion can encode a message when: the conversion knows how to encode messages with USING.name.
```
NielsZeilemaker, line 1295: I get why you changed it, but "using" isn't really a proper keyword.
Maybe using two keywords is a better approach.

Then you can get_conversion(message = X) or get_conversion(packet=Y).
Add an assert which checks if only one is passed and you're done.

boudewijn-tribler, line 1295: I agree that splitting it up in two variables would solve the naming problem, but it feels like calling get_conversion without the need to specify the keyword argument is more user friendly.

Would a more appropriate name be better? In which case I am open to suggestions.

NielsZeilemaker, line 1295: I guess calling get_conversion directly isn't really something a user should do.
If he does, choosing the correct parameter name isn't really that difficult.

boudewijn-tribler, line 1295: An alternative is to have several methods (get_default_conversion, get_conversion_for_packet, get_conversion_for_message).  This would be in-line with other get methods already in Dispersy, such as get_member, get_member_from_id.

whirm, line 1295: I like last @boudewijn-tribler's proposal slightly more than the others.

community.py
```diff
@@ -1279,51 +1279,81 @@ def get_members_from_id(self, mid):
...
+        if __debug__:
+            from .message import Message
+            assert isinstance(message, (Message, Message.Implementation)), type(message)
+
+        for conversion in reversed(self._conversions):
+            if conversion.can_encode_message(message):
+                return conversion
 
-    def add_conversion(self, conversion, default=False):
+        # for backwards compatibility we will raise a KeyError when no conversion for PREFIX is found (previously
```
NielsZeilemaker: Maybe change PREFIX with packet, otherwise I think it's a great improvement.

PR #103, opened 2013-07-22
Fixes undefined local variable.

boudewijn-tribler:


boudewijn-tribler:
Once merged, the Tribler next branch must be updated to include this fix.

whirm:
you should do it the other way around, fixes to tribler next and then merge to devel, or we have to manually cherry pick the fix commits.

synctext:
So we should have also a Dispersy "next" branch? Or just have Tribler Next point here?
Currently that Dispersy branch does not exist. That also would mean some extensions to Jenkins (Dispersy_test for: master,Devel,Next). Is that what you prefer Whirm?

whirm:
We can create the next branch the moment we need to stabilize Dispersy while developing new features. We are not in a rush right now.
Edit: I see we have the cleanup pull request pending, so maybe it is a good moment to create the next branch now.

Issue #102, opened 2013-07-22
variable 'community' referenced before assignment


synctext:
As reported by user VVS on [our forum](http://forum.tribler.org/viewtopic.php?f=4&t=3687&p=6974&sid=135550a75397b708f577ec36f5e9c2e2#p6974)

Traceback (most recent call last):
  File "/home/vvs/tribler-6.2.0-rc1/Tribler/dispersy/dispersy.py", line 2703, in store_update_forward
    messages[0].handle_callback(messages)
  File "/home/vvs/tribler-6.2.0-rc1/Tribler/dispersy/dispersy.py", line 2492, in on_introduction_response
    self.wan_address_vote(payload.destination_address, candidate)
  File "/home/vvs/tribler-6.2.0-rc1/Tribler/dispersy/dispersy.py", line 1066, in wan_address_vote
    for candidate in [candidate for candidate in community.candidates.itervalues() if candidate.wan_address == self._wan_address]:
UnboundLocalError: local variable 'community' referenced before assignment

NielsZeilemaker:
Maybe we can create a small unit-test for this bug, seems very trivial to trigger. Then we'll never run into this bug again.

boudewijn-tribler:
Fix: https://github.com/Tribler/dispersy/pull/103

I agree with the unit-test, I'm a bit cramped for time though so will postpone till after my vacation (unless someone else can build one).

NielsZeilemaker:
It's quite strange that this bug wasn't noticed during the new walking_nat tests:
http://jenkins.tribler.org/job/Test_community_walking_NAT/

As there actually symmetric-natted boxes are used, which should trigger this bug. Any clue why we're not seeing any exceptions there?

boudewijn-tribler:
No clue.  There must be a logical explanation though, perhaps Dispersy (#96) or the logger is eating the message?

boudewijn-tribler:
#105 solves the issue with Dispersy/logger hiding exceptions, should prevent exceptions like this from making it through our unit tests.

PR #104, opened 2013-07-22
Test for wan_voting

NielsZeilemaker:
No clue where to put this one, hence I create a new test file. #102

### DIFF
--- a/tests/test_nat_detection.py
+++ b/tests/test_nat_detection.py
@@ -1,10 +1,19 @@
 from .dispersytestclass import DispersyTestFunc, call_on_dispersy_thread
 from .debugcommunity.community import DebugCommunity
 
-class TestVoting(DispersyTestFunc):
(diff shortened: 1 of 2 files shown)


NielsZeilemaker:
The test is triggering the bug now, should work after Boudewijn's fix is merged.

NielsZeilemaker:
retest this please

boudewijn-tribler:
Thanks for making the test, but can you add a test description please.

NielsZeilemaker:
I will, any idea on a name for the test? I don't think there's a unit test for the nat type detection logic right? 
Maybe test_nat_detection.py?

boudewijn-tribler:
The only existing test that comes close is test_walker.py (as the nat detection occurs during the walker), but adding a new test makes more sense.  Your proposed name sounds good.

PR #106, opened 2013-07-24
RequestCache cleanup.

boudewijn-tribler:
- Identifiers are now unicode strings, these are used as keys in the
  RequestCache instance and as Callback id's.
- Cache now takes a unicode string as a parameter, this is the
  identifier.  Older code set the identifier (outside of the Cache
  constructor) once the cache was added to the RequestCache.
- NumberCache is used when, instead of a predefined unicode string,
  the identifier must contain a random number.  This is used by, for
  instance, the introduction-request message.  The random number is
  available from the 'Cache.number' property, while the
  'Cache.identifier' property still contains the unicode string.
  
  Note that this means that any older code that used to encode
  Cache.identifier into a request or response message must now use
  Cache.number instead.
- The static method Cache.create_identifier is used to generate a
  Cache specific unicode string.  Parameters can be supplied to add
  unique features, such as the number in NumberCache.
  
  Note that the generic method 'handle_missing_messages' needs to get
  the identifier from a message instance.  This is done using
  Cache.create_identifier_from_message, which in turn should call
  Cache.create_identifier with the appropriate parameters obtained
  from the message instance.  Code that uses 'handle_missing_messages'
  must ensure their Cache object supplies
  'create_identifier_from_message'
- RequestCache.add(CACHE) replaces RequestCache.set(ID, CACHE).
  RequestCache.add will return None when CACHE.identifier is already
  in use.
- RequestCache.claim no longer exists.  A valid (read unused)
  identifier can be found using RequestCache.has or repeated calls to
  RequestCache.add, while checking its return value.
- RequestCache.{replace, has, get} have had their parameters modified
  to reflect the introduction of the unicode string identifier.

### DIFF
--- a/requestcache.py
+++ b/requestcache.py
@@ -4,13 +4,28 @@ logger = logging.getLogger(__name__)
 from random import random
 
 
-def identifier_to_string(identifier):
-    return identifier.encode("HEX") if isinstance(identifier, str) else identifier
-
-
 class Cache(object):
-    timeout_delay = 10.0
-    cleanup_delay = 10.0
+
+    @staticmethod
+    def create_identifier():
+        raise NotImplementedError()
+
+    def __init__(self, identifier):
+        assert isinstance(identifier, unicode), type(identifier)
+        self._identifier = identifier
+
+    @property
+    def identifier(self):
+        assert isinstance(self._identifier, unicode), type(self._identifier)
+        return self._identifier
+
+    @property
+    def timeout_delay(self):
+        return 10.0
+
+    @property
+    def cleanup_delay(self):
+        return 10.0
 
     def on_timeout(self):
         raise NotImplementedError()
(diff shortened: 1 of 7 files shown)


requestcache.py
```diff
@@ -19,107 +34,172 @@ def on_cleanup(self):
...
+        assert isinstance(cache, Cache), type(cache)
+        assert isinstance(cache.identifier, unicode), type(cache.identifier)
+        assert isinstance(cache.timeout_delay, float), type(cache.timeout_delay)
+        assert cache.timeout_delay > 0.0, cache.timeout_delay
+
+        logger.debug("replace %s with %s", self._identifiers.get(cache.identifier), cache)
+        self._identifiers[cache.identifier] = cache
+        # 2013/07/23 Boudewijn: there appeared to be a bug with the delay parameter, it was using cleanup_delay instead
+        # of timeout_delay
+        self._callback.replace_register(cache.identifier, self._on_timeout, (cache,), delay=cache.timeout_delay)
```
boudewijn-tribler, line 118: The `replace` method is used from the private search community and it seems all the caches used there have `cleanup_delay=0.0`, so this may have been causing unnecessary timeouts.

tests/test_requestcache.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assertTrue(request_cache.has(cache.identifier))
+        self.assertEqual(request_cache.get(cache.identifier), cache)
+        # remove
+        self.assertEqual(request_cache.pop(cache.identifier), cache)
+
+        # has, get, and pop keep working because cache.cleanup_delay > 0.0.  Note that this is very strange and
+        # unexpected!  Why is it still there when it was just popped?
+        self.assertTrue(request_cache.has(cache.identifier))
+        self.assertEqual(request_cache.get(cache.identifier), cache)
+        self.assertEqual(request_cache.pop(cache.identifier), cache)
```
boudewijn-tribler, line 114: We should improve the `pop` behaviour, it doesn't make sense.
- We could remove the cleanup_delay entirely, although this would mean that we no longer have the ability to ensure that an identifier/number isn't used for some time.  It is currently used for the introduction request, adding 4.5 seconds before the identifier can be used again.  Without the cleanup_delay the randomness in the numbers will have to prevent clashes.
- We could add the identifier to a different dictionary or set, removing it from `RequestCache_identifiers`.  This would result in has, get, and pop no longer being able to find the identifier.  We would need a new way to see if an identifier is unclaimed though, for instance by introducing `is_used` or something similar that checks `_identifiers` and the new dict/set.
- I'm sure there are other options.

Anyway, removing the cleanup_delay entirely is the simplest change, and would result in a simplification of the RequestCache.  Are there cases where the delay is essential?

synctext, line 114: What I recall from discussion is that the pop() function not always behaves as expected.
The semantics of pop() are: return-and-remove-top-most-element().
By defition a pop() function does not do any search, it uses a stack approach with a LIFO policy.

So I suggest simply renaming this function or alter it's behavior.

boudewijn-tribler, line 114: The RequestCache behaves like a dictionary (i.e. stores Caches associated with their keys, or identifiers in our case).  Hence the pop() behavior that we should mimic is the behavior described by python dictionaries (see: http://docs.python.org/2/library/stdtypes.html#dict.pop)

In both the python dict.pop and a stack based LIFO pop the item that is 'popped' should be removed from the cache.  That our RequestCache allows multiple pops of the same key/identifier is not only in conflict with its name, but it _can_ also cause bugs during runtime that malicious peers _could perhaps_ exploit.

Hence, unfortunately we can not just rename the function (which would be fast and easy).  Instead we will have to fix its behavior to conform to what dict.pop() would do, i.e. actually remove the item when popped.

boudewijn-tribler:
Once this pull is accepted I will make a pull request to update Tribler (code works on my machine).  As for the discussion with pop, this will not be modified in this pull request.  Fixing pop will result in different behavior while this pull request keeps the behavior the same, making it easier to verify the correctness.  Hence I want to fix the pop separately.

If no one objects this can be merged.

PR #107, opened 2013-07-26
Fixes transitions between connection types.

boudewijn-tribler:
Votes from external peers will influence our WAN address and our
connection type.  New unit tests are included to test Dispersy making
the correct assessment, event when votes change over time.

The code has been simplified, as the previous code was unable to
correctly change connection type.  Also, updating the LAN address was
introduced.


whirm:
ok to test

PR #115, opened 2013-08-15
Using Python's abc to handle abstract methods and properties.

boudewijn-tribler:
Since Python 2.6 the abc module provides a way to define abstract
methods and properties.  Before this the convention was to raise a
NotImplementedError.

Using abc will ensure an error is given when an object is instantiated
instead of waiting for the abstract method/property to be called.


boudewijn-tribler:
retest this please

PR #121, opened 2013-09-05
Fixes and improvement

boudewijn-tribler:
- New file: logger.py
  By importing logger.py and using the get_logger it provides we can ensure that every logger is correctly configured.  This allows us to add data to each log entry, such as the hostname, which allows for easier debugging when running large DAS experiments.
- Reintroduced random ordered bloom filter synchronization
  Experiments make it clear that random selection of packets to sync is more efficient in scenarios with relatively few 'seeding' nodes.
- A few updates to code that was not yet up-to-date with the latest Dispersy
- Changed several logger.warning into logger.debug
  Warning level should not be used for something that is standard in every DAS4 experiment run


whirm:
The logging part looks OK to me.
@egbertbouman: You are the one working the most with Tribler, anything to object with this PR? (if not, just merge it. thx)

PR #122, opened 2013-09-11
Add bloom creation wall time counter to statistics.

whirm:


boudewijn-tribler:
Conclusion from a discussion on IRC: we will look into using `attach_runtime_statistics()` and providing its results to Dispersy.statistics.

whirm:
Looking at your thing we see it's not the same, so Ill reopen the pull request.

boudewijn-tribler:
Can you explain why the decorator won't work for what you want?

NielsZeilemaker:
I would like to integrate this into the default statistics to be able to see the time spent creating bloom filters whenever I run an experiment. Moreover, I think this statistics should be included in all experiments we perform daily with Dispersy to be able to check if we made bloom filter creation worse/better.

Maybe we can even extend the statistic to also include the walltime spent on replying to bloom filters. This would allow us to diagnose potential problems even better.

boudewijn-tribler:
Yes, but this was already clear.  So why not use the existing decorator?  Surely adding a decorator to all interesting methods is a cleaner solution than adding timers throughout the code.

NielsZeilemaker:
But then all experiments should run with this decorator enabled + johan should run it with this decorator enabled etc.

boudewijn-tribler:
Correct.  That isn't a problem for the experiments, and I am sure Johan will be able to remember it as well.

NielsZeilemaker:
If we enable it for all experiments, then we get all sql-statements logged as well. That might be a bit much.

boudewijn-tribler:
It would give the same benefit as you described earlier, i.e. allowing to see if bloom filter creation gets worse or better, only now applied to sql-statements (and any other generator we decide to add).

boudewijn-tribler:
Egbert is looking into visualizing the statistics, when he is done (and the structure of the data is known) we will push the decorator 'runtime statistics' to the Dispersy and Tribler repro's.

I suggest we close this pull request.

NielsZeilemaker:
I would really like to see that this decorator is then enabled by default, or at least is dependent on the are_debug_statistics_enabled() method. If it is, then we can at least easily toggle it without having to restart Tribler etc.

boudewijn-tribler:
Yea, the current idea is to always have it enabled.  In fact, the '--enable-runtime-statistics' check will be removed.

NielsZeilemaker:
Superseded by #146

PR #125, opened 2013-09-13
Remove tunnel prefex before converting to meta.

### LINKED ISSUE #124: Calls to convert_packet_to_meta_message() might pass tunnel-prefixed packets
Line `endpoint.py:212` calls `convert_packet_to_meta_message()` which expects a packet **without** the optional Swift tunnel prefix. If this gets called with a tunnel-prefixed message it will result in a `get_community()` call with the wrong CID. The tunnel prefix **is not removed** before the call.

`convert_packet_to_meta_message()` dumps 2 bytes (Dispersy version field) from the beginning of the packet, to get the CID, then calls `get_community(cid)` (which might have side-effects). Assuming CID is `deadbeaf`:
- without tunnel prefix: `0001deadbeaf` => `deadbeaf` (correct)
- with tunnel prefix: `ffffffff0001deadbeaf` => `ffff0001dead`

This might also apply for: `endpoint.py:377`, `endpoint.py:396`, and `endpoint.py:142`.

For example, this affects `TrackerDispersy`, which tracks communities when its `get_community()` is called.

boudewijn-tribler:
Fixes #124.

### DIFF
--- a/tests/test_overlay.py
+++ b/tests/test_overlay.py
@@ -159,23 +159,50 @@ class TestOverlay(DispersyTestFunc):
         summary.debug("\n%s", pformat(self._dispersy.statistics.get_dict()))
 
         # write graph statistics
-        handle = open("%s_connections.txt" % cid_hex, "w+")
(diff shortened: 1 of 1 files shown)


boudewijn-tribler:
No Idea why I closed this.

boudewijn-tribler:
Apparently it was already merged

PR #131, opened 2013-10-02
Bug with starting Dispersy and code cleanups.

boudewijn-tribler:
- Fixes a bug when starting from dispersy/main.py and dispersy/tracker.py.
- Removed some obsolete code.
- Providing cleaner feedback on Dispersy.start()
- A few other code/debug-output improvements.


boudewijn-tribler:
Needs a unit test to ensure that starting the tracker doesn't break again in the future.

PR #135, opened 2013-10-14
Non-blocking DNS lookup.

boudewijn-tribler:
http://jenkins.tribler.org/job/Test_community_walking_NAT/176/console

The Jenkins job above blocks a very long time when DNS lookup fails.  This code ensures that the Dispersy thread doesn't block for longer than a given timeout, allowing the Jenkins job to quit more timely.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,7 +1,10 @@
-import os
+from os import path
 from socket import gethostbyname
+from threading import Thread, Lock, Event
 
 from .candidate import BootstrapCandidate
+from .logger import get_logger
+logger = get_logger(__name__)
 
 _trackers = [(u"dispersy1.tribler.org", 6421),
              (u"dispersy2.tribler.org", 6422),
@@ -21,8 +24,6 @@ _trackers = [(u"dispersy1.tribler.org", 6421),
              (u"dispersy7b.tribler.org", 6427),
              (u"dispersy8b.tribler.org", 6428)]
 
-# _trackers = [(u"kayapo.tribler.org", 6431)]
-
 
 def get_bootstrap_hosts(working_directory):
     """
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -46,7 +47,7 @@ def get_bootstrap_hosts(working_directory):
         return _trackers
 
 
-def get_bootstrap_candidates(dispersy):
+def get_bootstrap_candidates(dispersy, timeout=1.0):
```
NielsZeilemaker, line 50: Using 1 second as a default timeout might be overly strict. I guess using a more sane value like 30 seconds would be better.

NielsZeilemaker:
Moreover, this change renders the following line in dispersy.py completely useless:

``` python
        bootstrap_candidates = get_bootstrap_candidates(self)
        if not all(bootstrap_candidates):
            self._callback.register(self._retry_bootstrap_candidates)
```

bootstrap.py
boudewijn-tribler, line 50: I don't think calling this method should block for 30 seconds, especially since it can be called multiple times whenever there are dns issues (as we are now seeing on the DAS2).

Perhaps it is better to rewrite it, giving it a callback when its finished.

NielsZeilemaker, line 50: I agree, it shouldn't be blocking. Rewriting it in combination with the dispersy.py lines is the only option.

NielsZeilemaker, line 50: Btw, as I said below. Dispersy will not retry/call this method multiple times as results will not contain empty positions for timedout dns requests.

boudewijn-tribler, line 50: Yea, that was indeed a bug.  I'll have to rewrite the entire retry method to take into account the non-blocking callback

boudewijn-tribler, line 50: See #138.

PR #138, opened 2013-10-17
Blocking gethostbyname

boudewijn-tribler:
The addresses of BootstrapCandidate instances are resolved on a separate thread, on success or timeout a callback is performed.

The goal is to ensure Tribler doesn't block when the dns resolving is having problems (slow or unresponsive dns servers, or simply an address that is no longer valid).  However, for the Tribler search community to work correctly the 'fast bootstrap' must have bootstrap candidates.  Therefore the Dispersy.start() will block at most timeout=10 seconds until 50% of the bootstrap servers are resolved.

Note that Dispersy will keep trying to resolve (between 300 and 600 seconds intervals, depending on the failure reason) until it has resolved all addresses.  Once an address is resolved it is not resolved again.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -1,100 +1,158 @@
-from os import path
 from socket import gethostbyname
-from threading import Thread, Lock, Event
+from threading import Lock, Event
+from random import shuffle
 
+from .callback import Callback
 from .candidate import BootstrapCandidate
 from .logger import get_logger
 logger = get_logger(__name__)
 
-_trackers = [(u"dispersy1.tribler.org", 6421),
-             (u"dispersy2.tribler.org", 6422),
-             (u"dispersy3.tribler.org", 6423),
-             (u"dispersy4.tribler.org", 6424),
-             (u"dispersy5.tribler.org", 6425),
-             (u"dispersy6.tribler.org", 6426),
-             (u"dispersy7.tribler.org", 6427),
-             (u"dispersy8.tribler.org", 6428),
+_DEFAULT_ADDRESSES = ((u"dispersy1.tribler.org", 6421),
+                      (u"dispersy2.tribler.org", 6422),
+                      (u"dispersy3.tribler.org", 6423),
+                      (u"dispersy4.tribler.org", 6424),
+                      (u"dispersy5.tribler.org", 6425),
+                      (u"dispersy6.tribler.org", 6426),
+                      (u"dispersy7.tribler.org", 6427),
+                      (u"dispersy8.tribler.org", 6428),
 
-             (u"dispersy1b.tribler.org", 6421),
-             (u"dispersy2b.tribler.org", 6422),
-             (u"dispersy3b.tribler.org", 6423),
-             (u"dispersy4b.tribler.org", 6424),
-             (u"dispersy5b.tribler.org", 6425),
-             (u"dispersy6b.tribler.org", 6426),
-             (u"dispersy7b.tribler.org", 6427),
-             (u"dispersy8b.tribler.org", 6428)]
-
-
-def get_bootstrap_hosts(working_directory):
-    """
-    Reads WORKING_DIRECTORY/bootstraptribler.txt and returns the hosts therein, otherwise it
-    returns _TRACKERS.
-    """
-    trackers = []
-    filename = path.join(working_directory, "bootstraptribler.txt")
-    try:
-        for line in open(filename, "r"):
-            line = line.strip()
-            if not line.startswith("#"):
-                host, port = line.split()
-                trackers.append((host.decode("UTF-8"), int(port)))
-    except:
-        pass
-
-    if trackers:
-        return trackers
-    else:
-        return _trackers
+                      (u"dispersy1b.tribler.org", 6421),
+                      (u"dispersy2b.tribler.org", 6422),
+                      (u"dispersy3b.tribler.org", 6423),
+                      (u"dispersy4b.tribler.org", 6424),
+                      (u"dispersy5b.tribler.org", 6425),
+                      (u"dispersy6b.tribler.org", 6426),
+                      (u"dispersy7b.tribler.org", 6427),
+                      (u"dispersy8b.tribler.org", 6428))
 
 
-def get_bootstrap_candidates(dispersy, timeout=1.0):
-    """
-    Returns a list with all known bootstrap peers.
+class Bootstrap(object):
 
-    Bootstrap peers are retrieved from WORKING_DIRECTORY/bootstraptribler.txt if it exits.
-    Otherwise it is created using the trackers defined in _TRACKERS.
+    @staticmethod
+    def load_addresses_from_file(filename):
+        """
+        Reads FILENAME and returns the hosts therein, otherwise returns an empty list.
+        """
+        addresses = []
+        try:
+            for line in open(filename, "r"):
+                line = line.strip()
+                if not line.startswith("#"):
+                    host, port = line.split()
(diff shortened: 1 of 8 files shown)


bootstrap.py
```diff
@@ -1,100 +1,160 @@
...
+        assert timeout > 0.0, timeout
+
+        if self.are_resolved:
+            self._callback.register(func, (True,))
+
+        else:
+            self._thread_counter += 1
+
+            # start a new thread (using Callback to ensure the thread is named properly)
+            thread = Callback("Get-Bootstrap-Candidates-%d" % self._thread_counter)
```
synctext, line 122: Is here an entire event handler started to handle a single DNS lookup event?

callback.py
```diff
@@ -670,6 +678,9 @@ def _one_task(self):
 
     @attach_profiler
     def loop(self):
+        # set thread name (visible from ps and top)
+        set_name(self._name[:16])
```
synctext, line 682: Is the thread name equal to something like "DNS_Dispersy1.tribler.org ?
So we can see nicely what is hanging for 60 or something seconds?

dispersy.py
```diff
@@ -438,24 +436,53 @@ def _guess_lan_address(interfaces, default=None):
...
+            for community in self._communities.itervalues():
+                community.update_bootstrap_candidates(self._bootstrap_candidates.itervalues())
+
+            if success:
                 logger.debug("resolved all bootstrap addresses")
-                self._bootstrap_candidates = dict((candidate.sock_addr, candidate) for candidate in candidates if candidate)
-                break
+
+            else:
+                delay = 0.0 if container["attempt_counter"] == 1 else 300.0
```
synctext, line 466: What are we doing here every 5 minutes?

dispersy.py
```diff
@@ -4447,22 +4474,35 @@ def _commit_now(self):
...
+        3. opens database
+        4. opens endpoint
         """
 
         assert not self._callback.is_running, "Must be called before callback.start()"
+        assert isinstance(timeout, float), type(timeout)
+        assert timeout >= 0.0, timeout
+
+        def bootstrap_progress(_, resolved, total):
+            # we are happy starting with just 50% of the available BootstrapCandidates,
```
synctext, line 4494: Is there proof that 50% speeds up Dispersy?
What is this policy based on? Why not 60%?

If DNS lookup is broken we have a timeout to deal with that.
So can be simplified.

synctext:
As you've seen above I've done a bit of a read-through of the code. 

My quick general comments:
It is a bit complex and not easy to read. I prefer the most basic approach: If a tracker lookup timeouts after 60 seconds it should be discarded and we move on with a smaller list and Tribler starts further. The multiple timeouts, single-use event handler (if I read the code correct), Lock() and 50% threshold is not required for Tribler I believe.

Can you please see if you can also make a better patch without the 4 things and with minimal number of changes to 1 source file?

btw the term non-blocking is reserved for something different, this is a parallel asynchronous DNS lookup.

bootstrap.py
```diff
@@ -1,100 +1,160 @@
...
-             (u"dispersy8.tribler.org", 6428),
-
-             (u"dispersy1b.tribler.org", 6421),
-             (u"dispersy2b.tribler.org", 6422),
-             (u"dispersy3b.tribler.org", 6423),
-             (u"dispersy4b.tribler.org", 6424),
-             (u"dispersy5b.tribler.org", 6425),
-             (u"dispersy6b.tribler.org", 6426),
-             (u"dispersy7b.tribler.org", 6427),
-             (u"dispersy8b.tribler.org", 6428)]
```
synctext, line 25: 16 trackers on 1 DNS domain is overkill.

Please alter to 8 central trackers on 3 DNS domains/IP ranges.
So, 1 cs.vu.nl (or leaseweb if we lost that Arno machine access).
1 on TUDelft directly
and other 6 on tribler.org domain, as above.

boudewijn-tribler, line 25: About a year ago we discussed this and decided to increase the
number of entries from 8 to 16 because we wanted the possibility
to add more physical machines in the future.  We choose this
setup because currently we only use 4 different machines.  Do we
no longer want this?

I agree that we should use more DNS domains.  But we need to use
domains that can be changed when a tracker is moved or removed.  
Can we do this with the vu domain once cs.vu.nl goes offline?  Do 
we have direct or indirect access to a third domain?

bootstrap.py
boudewijn-tribler, line 122: No.  One event handler will perform as many DNS lookups as
possible until it exceeds its timeout.

We use an event handler to re-use the code that sets the thread
name.  Not setting the thread name with the `prctl` library
would inherit the thread name from its parent, i.e. Dispersy.

callback.py
boudewijn-tribler, line 682: Yes, this will allow us to see these threads in top and ps.

The thread name is the parameter given when the thread is created:
`thread = Callback("Get-Bootstrap-Candidates-%d" % self._thread_counter)`
So that would be the first 16 characters
of "Get-Bootstrap-Candidates-%d".  The 16 character limit is
imposed by `prctl`.

dispersy.py
boudewijn-tribler, line 466: This will retry the DNS lookup when all of the DNS entries were
not yet resolved.  I did not want to change this behavior as it 
was doing this before.

dispersy.py
boudewijn-tribler, line 4494: Only relying on the timeout will slow down the starting Dispersy
by `timeout` seconds.  Requiring fewer DNS lookups to succeed
results in faster startup.  This was tested by adding numerous
invalid trackers such as 'invalid.domain.name42.eu'.

The percentage effectively comes down to how many DNS lookups can
fail before it starts slowing down Dispersy.start().  50% will
currently leave us with a minimum of 4 out of 8 trackers, and
bootstrapping generally takes only two trackers.  So the value
seems reasonable.

boudewijn-tribler:
It is good to see you are taking an interest in this patch!  I am
aware that it adds complexity to the code.  Unfortunately the
problem it tries to solve is also complex.

It can only be simplified when we make the problem simpler,
- do we mind Dispersy.start() waiting its full timeout when one
  or more DNS lookups fail?
- do we mind the fast bootstrap mechanism used by the
  SearchCommunity to fail when one or more DNS lookups fail?

We can simplify the code when the answer to one or both of these
questions is _no_.

bootstrap.py
synctext, line 25: 8 different machines in 3 different buildings build be ideal. 16 is overkill.
As said, please use 1 TUDelft hard coded server DNS entry.
We can alter that DNS entry or assign it to a different machine. Thus:

Dispersy1.tribler.org ...
Dispersy6.tribler.org
om.cs.vu.nl
superpeer9.das2.ewi.tudelft.nl

synctext:
- We don't mind no. 1.
- "do we mind the fast bootstrap mechanism used by the SearchCommunity to fail when one or more DNS lookups fail?"

Why would that fail? With less then 8 bootstrap servers, we can still boot quickly.

Please redo this patch with minimal complexity:
- 8 threads; not full-blown event handlers (for 8 bootstrap servers), 15 second timeout
- Dispersy Blocks Boostrapping.
- After 15 seconds: lookups that fail, scheduled for retry after 300seconds.

So if DNS is not available and all looksups fail we simply retry everything in 300seconds and Search community will bootstrap then.

boudewijn-tribler:
Ok, I'll prepare a simpler version asap.

PR #143, opened 2013-10-28
Don't fail when 9/10 attempts succeed.

boudewijn-tribler:


boudewijn-tribler:
Segmentation fault after the unit tests finished successfully

PR #146, opened 2013-10-31
Provide runtime statistics to statistics.py

boudewijn-tribler:
This is a requirement for https://github.com/Tribler/tribler/pull/294

### DIFF
--- a/decorator.py
+++ b/decorator.py
@@ -1,10 +1,12 @@
 from atexit import register as atexit_register
 from cProfile import Profile
 from collections import defaultdict
+from functools import wraps
 from hashlib import sha1
 from thread import get_ident
 from threading import current_thread
 from time import time
+import logging
 import sys
 
 from .logger import get_logger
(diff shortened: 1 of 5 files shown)


NielsZeilemaker:
Seems fine, maybe add the runtime statistics to some other cpu intensive tasks as well.
Creating signatures? Time it takes to create a message of a given type?

boudewijn-tribler:
Wrapping the Conversion.encode_message and Conversion.decode_message methods will give us creating and verifying of signatures and the byte packing stuff per message message.  I'll update the code so you can have a look.

NielsZeilemaker:
Maybe wrap the encode/decode_member_authentication and encode/decode_double_member_authentications to get a more clear understanding of just the signing part of messages.

boudewijn-tribler:
I did and all it shows is that approximately 90% of the decode is with the crypto verify and approximately 80% of the encode is in the signing part.

Of course these percentages won't be the same for every message, but I remember measuring this about a year ago with the same results.  So before I push this commit, do you still think this information is interesting enough to have it measured all the time?

NielsZeilemaker:
Yes, because if we measure it we can see if these assumptions continue to hold.

boudewijn-tribler:
Segmentation fault occurring after the tests run successfully...  Elric, can you take a look at this please?

decorator.py
```diff
@@ -71,64 +73,101 @@ def helper(*args, **kargs):
...
-           foo(2)
-
-        After running the above example, the statistics will show that:
-        - 'foo bar=1 moo=milk' was called once
-        - 'foo bar=2 moo=milk' was called twice
-        """
-        def helper(func):
-            return func
-        return helper
+    Runtime statistics will only be collected when sys.argv contains '--runtime-statistics'.
```
NielsZeilemaker: Not true

Issue #152, opened 2013-11-06
Dispersy.database should not log errors


NielsZeilemaker:
When using database.py I ran into the following problem.

I want to insert or update a row into a table, but in order to improve the performance I don't want to first perform a select to see if the row exists and then do a insert or update accordingly, but I just perform the insert and handle the exception when the row existed.

This works fine, however the execute, executemany, and executescript methods have already logged the "row existed/primary key violation thingy" as an exception.
Can we remove this logger.exception calls from database.py?

boudewijn-tribler:
Yes we can.  These cases are already raising the exception allowing it to be logged higher up the chain (if desired) or ignored (as Niels needs).

NielsZeilemaker:
Great, if you could remove the try excepts, then my log will be significantly cleaner.

boudewijn-tribler:
Yea, I can imagine.

PR #153, opened 2013-11-06
Exceptions that are raised should not be logged.

boudewijn-tribler:
This will allow people to ignore statements that they know can fail.

See #152


NielsZeilemaker:
Seems fine, thanks

PR #154, opened 2013-11-07
Fixes API inconsistency.

boudewijn-tribler:
Member.sign and Member.verify used the OFFSET and LENGTH parameters
differently.  They now both use the one Member.verify used, i.e. LENGTH
is the number of bytes, starting at offset, that is either verified or
signed.

In addition to this change, whenever there is insufficient DATA
available, i.e. len(DATA) < OFFSET + LENGTH, then Member.verify will
return False and Member.sign will raise a ValueError.

### DIFF
--- a/crypto.py
+++ b/crypto.py
@@ -157,6 +157,7 @@ def ec_sign(ec, digest):
     """
     Returns the signature of DIGEST made using EC.
     """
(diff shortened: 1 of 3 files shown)


NielsZeilemaker:
I guess it passes the unit-tests right? If so, then please merge it.

PR #159, opened 2013-11-13
Process guard killing Dispersy it is not an error.

boudewijn-tribler:


NielsZeilemaker:
Which process guard is this?

boudewijn-tribler:
Discussed in person.

PR #160, opened 2013-11-13
Externalized sql query for Bloomfilters, Introduced NoDefBinaryConversion

NielsZeilemaker:
- The new externalized sql query handling a Bloomfilter allows a developer to determine which packets should be tested against an incomming Bloomfilter.
- The NoDefBinaryConversion allows a developer to extend a conversion object which does not define all default Dispersy messages. Usefull when you want to only define a couple messages in a conversion object and let other conversion classes to overwrite any of the default Dispersy messages.

### DIFF
--- a/callback.py
+++ b/callback.py
@@ -300,7 +300,7 @@ class Callback(object):
             else:
                 heappush(self._requests,
                          (delay + time(),
-                          -priority,
+                          - priority,
                           id_,
                           (call, args + (id_,) if include_id else args, {} if kargs is None else kargs),
                           None if callback is None else (callback, callback_args, {} if callback_kargs is None else callback_kargs)))
@@ -363,7 +363,7 @@ class Callback(object):
                     else:
                         heappush(self._requests,
                                  (delay + time(),
-                                  -priority,
+                                  - priority,
                                   id_,
                                   (call, args + (id_,) if include_id else args, {} if kargs is None else kargs),
                                   None if callback is None else (callback, callback_args, {} if callback_kargs is None else callback_kargs)))
@@ -418,7 +418,7 @@ class Callback(object):
             else:
                 heappush(self._requests,
                          (delay + time(),
-                          -priority,
+                          - priority,
                           id_,
                           (call, args + (id_,) if include_id else args, {} if kargs is None else kargs),
                           None if callback is None else (callback, callback_args, {} if callback_kargs is None else callback_kargs)))
@@ -661,6 +661,9 @@ class Callback(object):
                 self._call_exception_handlers(exception, True)
 
             except Exception as exception:
+                from traceback import print_exc
+                print_exc()
+
                 if callback:
                     with self._lock:
                         heappush(self._expired, (priority, actual_time, root_id, (callback[0], (exception,) + callback[1], callback[2]), None))
--- a/callback.py
+++ b/callback.py
@@ -661,9 +661,6 @@ class Callback(object):
                 self._call_exception_handlers(exception, True)
 
             except Exception as exception:
-                from traceback import print_exc
-                print_exc()
-
                 if callback:
                     with self._lock:
                         heappush(self._expired, (priority, actual_time, root_id, (callback[0], (exception,) + callback[1], callback[2]), None))
--- a/conversion.py
+++ b/conversion.py
@@ -135,8 +135,7 @@ class Conversion(object):
     def __repr__(self):
         return str(self)
 
-
-class BinaryConversion(Conversion):
+class NoDefBinaryConversion(Conversion):
 
     """
     On-The-Wire binary version
@@ -144,6 +143,7 @@ class BinaryConversion(Conversion):
     This conversion is intended to be as space efficient as possible.
     All data is encoded in a binary form.
     """
+
     class Placeholder(object):
         __slots__ = ["candidate", "meta", "offset", "data", "authentication", "resolution", "first_signature_offset", "destination", "distribution", "payload", "verify", "allow_empty_signature"]
 
@@ -184,7 +184,7 @@ class BinaryConversion(Conversion):
             self.destination = destination
             self.payload = payload
 
-    def __init__(self, community, community_version):
+    def __init__(self, community, commnity_version):
         Conversion.__init__(self, community, "\x00", community_version)
 
         self._struct_B = Struct(">B")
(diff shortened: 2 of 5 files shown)


NielsZeilemaker:
retest this please

dispersy.py
```diff
@@ -2476,10 +2473,49 @@ def on_introduction_request(self, messages):
...
+                    if byte_limit <= 0:
+                        logger.debug("bandwidth throttle")
+                        break
+
+                if packets:
+                    logger.debug("syncing %d packets (%d bytes) to %s", len(packets), sum(len(packet) for packet in packets), message.candidate)
+                    self._statistics.dict_inc(self._statistics.outgoing, u"-sync-", len(packets))
+                    self._endpoint.send([message.candidate], packets)
+
+    def _get_packets_for_bloomfilters(self, community, requests, include_inactive=True):
```
NielsZeilemaker, line 2515: We could move this method to communty.py, I didn't do this as some other methods are planned to be moved as well.
Alternatively, we could add a small method in community.py which calls this method but call the community._get_packets_for_bloomfilters from within the on_intro_request and create_intro_request methods.

NielsZeilemaker:
@boudewijn-tribler any feedback on this?

conversion.py
```diff
@@ -1443,6 +1405,52 @@ def decode_message(self, candidate, data, verify=True):
...
         return "<%s %s%s [%s]>" % (self.__class__.__name__, self.dispersy_version.encode("HEX"), self.community_version.encode("HEX"), ", ".join(self._encode_message_map.iterkeys()))
 
+class BinaryConversion(NoDefBinaryConversion):
+
+    """
+    Extends NoDefBinaryConversion and will define all standard dispersy messages
+    """
+
+    def __init__(self, community, community_version):
+        NoDefBinaryConversion.__init__(self, community, community_version)
```
boudewijn-tribler: Can you use `super(...)` instead.

dispersy.py
boudewijn-tribler, line 2515: It would be simpler if this method returned the packets associated to a single request instead, then simply call this method once for each message that has a sync enabled.  The time_low, time_high, etc, don't need to be passed as arguments, they should be computed in `_get_packets_for_bloomfiler` directly (similar to all other variables needed when creating the SQL statement).

boudewijn-tribler, line 2515: This would also simplify the calling code since creating and iterating over `messages_with_sync` is no longer needed, i.e. we only need to iterator over `messages`.

dispersy.py
```diff
@@ -2517,44 +2553,19 @@ def get_sub_select(meta):
...
-                                          long(payload.offset),
-                                          long(payload.modulo)))
-                logger.debug("%s", sql_arguments)
-
-                packets = []
-                generator = ((str(packet),) for packet, in self._database.execute(sql, sql_arguments))
+        for message, time_low, time_high, offset, modulo in requests:
+            sql_arguments = []
+            for meta in meta_messages:
+                if include_inactive:
```
boudewijn-tribler, line 2575: Please document `include_inactive` in a docstring for this method

dispersy.py
NielsZeilemaker, line 2515: We can't as this method is also used to verify a Bloomfilter which was created using one of the claim_bloomfilter_\* methods.

Moreover, passing a single request will cause us to rebuild the meta_messages list and sqlstatement for each request. This you told me was something which needed to be avoided.

boudewijn-tribler, line 2515: I see.  So be it then.

dispersy.py
```diff
@@ -2476,10 +2473,63 @@ def on_introduction_request(self, messages):
...
+                    self._endpoint.send([message.candidate], packets)
+
+    def _get_packets_for_bloomfilters(self, community, requests, include_inactive=True):
+        """
+        Return all packets matching a Bloomfilter request
+        
+        @param community: The community wherein all requests were received
+        @type messages: [Community]
+
+        @param request: A list of requests, each of them being a tuple consisting of the request,
```
boudewijn-tribler: request -> requests

and perhaps also add that it yields (request, packets) tuples, not just packets.

conversion.py
```diff
@@ -1443,6 +1405,52 @@ def decode_message(self, candidate, data, verify=True):
...
         return "<%s %s%s [%s]>" % (self.__class__.__name__, self.dispersy_version.encode("HEX"), self.community_version.encode("HEX"), ", ".join(self._encode_message_map.iterkeys()))
 
+class BinaryConversion(NoDefBinaryConversion):
+
+    """
+    Extends NoDefBinaryConversion and will define all standard dispersy messages
+    """
+
+    def __init__(self, community, community_version):
+        super(BinaryConversion).__init__(community, community_version)
```
boudewijn-tribler: super(BinaryConversion, **self**)

callback.py
```diff
@@ -661,6 +661,9 @@ def _one_task(self):
                 self._call_exception_handlers(exception, True)
 
             except Exception as exception:
+                from traceback import print_exc
+                print_exc()
+
```
boudewijn-tribler: We should use `logger.exception` instead of print_exc.

But in this case it isn't needed at all, the callstack should be printed in the `_call_exception_handlers` method by line `logger.warning("keep running regardless of exception [%s]", exception)`.

This `logger.warning` statement should be renamed to `logger.exception` to get the callstack.  This has already been done on the Bartercast3 branch which should be merged into devel next week.

requestcache.py
```diff
@@ -181,7 +181,8 @@ def _on_timeout(self, cache):
         if cache.cleanup_delay:
             self._callback.replace_register(cache.identifier, self._on_cleanup, (cache,), delay=cache.cleanup_delay)
 
-        else:
+        # the on_timeout call could have already removed the identifier from the cache using pop
+        elif cache.identifier in self._identifiers:
```
boudewijn-tribler, line 185: This should not needed.  The pop ensures that `_on_timeout` is not called (either by using `callback.unregister` or `callback.replace_register`).

Can you tell me how to reproduce the bug that requires this fix?

requestcache.py
```diff
@@ -201,5 +202,8 @@ def _on_cleanup(self, cache):
 
         logger.debug("cleanup on %s", cache)
         cache.on_cleanup()
-        del self._identifiers[cache.identifier]
+
+        # the on_cleanup call could have already removed the identifier from the cache using pop
+        if cache.identifier in self._identifiers:
+            del self._identifiers[cache.identifier]
```
boudewijn-tribler, line 208: Same as https://github.com/Tribler/dispersy/pull/160#discussion_r8156553

callback.py
boudewijn-tribler: Changing `logger.warning` into `logger.exception` (as described above) is not on the Bartercast3 branch, it is in a pull request that is currently waiting.  See https://github.com/boudewijn-tribler/dispersy/commit/ac3da3afb208ec18326bd2255f3b05bced9472b9

requestcache.py
boudewijn-tribler, line 185: After a small conversation with @NielsZeilemaker: The use-case is when `RequestCache.pop()` is called from within the `Cache.on_timeout` or `Cache.on_cleanup` handler.  This fix will solve the problem that would otherwise occur.

PR #170, opened 2013-12-02
Fixes typo.

boudewijn-tribler:
Bug occurred only when the 'runtime-statistics' console logger is enabled.


boudewijn-tribler:
I'm also OK with removing the entire console output logger (i.e. the entire '_output_runtime_statistics' function) since most of the time we will be able to get the same information directly from the Tribler debug screen or from running the profiler.

NielsZeilemaker:
I guess you're the only one using this functionality, so if you don't need it anymore feel free to remove it.

boudewijn-tribler:
OK, less is more.

PR #172, opened 2013-12-05
Simplified bootstrap candidate DNS resolver.

boudewijn-tribler:


NielsZeilemaker:
I still don't like that its using a new callback instance to do the dnsresolve. This class is much to complex, a Thread with a target = method would suffice.
Moreover, the event timeout thingy doesn't really work. You should use a signal http://docs.python.org/2/library/signal.html to be able to break of out the blocking dns resolve call. An example can be found here: https://mail.python.org/pipermail/python-list/2009-November/559662.html

boudewijn-tribler:
I am using Callback because I want to also set the thread a name through prctl.set_name, which is already done in callback.py, so I can reuse this.  We disagree on the _to much_ complexity of the Callback class.

As for using `Event`, I know it won't interrupt the blocking dns resolve call.  Effectively the thread will remain alive until the dns resolve call times out by itself.  Since the thread is a daemon it will not cause Tribler to block, for example, during shutdown.

Using signals isn't a good option since we are using (many) threads and the main Dispersy thread isn't always the main thread.  From the site you link: _Some care must be taken if both signals and threads are used in the same program. The fundamental thing to remember in using signals and threads simultaneously is: always perform signal() operations in the main thread of execution._  I am sure that we can make signals work, but I believe an Event is simpler and perhaps safer.

NielsZeilemaker:
If you want to keep using callback, and not use signals then I would suggest to remove the event completely and switch to using tread.call if a timeout is passed. 

Callback.call will actually block for at most timeout seconds, wherein your current version does not.

boudewijn-tribler:
My current version _does_ block for at most timeout seconds because of the `thread.join(timeout)` statement.

NielsZeilemaker:
But you are rebuilding the functionality provided by Callback.call

boudewijn-tribler:
That is true, I've pushes a few changes.

NielsZeilemaker:
Maybe remove the self._thread_counter ?

boudewijn-tribler:
The counter is intended to distinguish between multiple threads running at the same time.  For example, when you call `Bootstrap.resolve(...)` a 3 times in a row, you will end up with 3 threads running simultaneously (especially when the dns resolve takes a long time)

Having a unique name should make it clear that it is a different thread and will also allow us to see how many have been created so far.

But... I just remembered that `prctl.set_name` will only show 16 characters, so with the current thread name `"Get-Bootstrap-Candidates-%d"` the above arguments don't hold up.  So I can either remove the counter, or make the name shorted (i.e. `"Resolve-DNS-%d"`) to get the number to fit within the first 16 characters.  Do you have a preference?

NielsZeilemaker:
I would prefer shortening it

boudewijn-tribler:
I made it `"Resolve-%d"`.  Adding `-DNS` would 'only' allow three characters for the number.

NielsZeilemaker:
Seems fine now, do you want to have elric review it as well? If not, feel free to merge it.

boudewijn-tribler:
OK, thanks.  I'll go over it with Elric (hopefully) today.

boudewijn-tribler:
Actually found two bugs yesterday.

Also cleaned up logging in callback.  Making is easier to avoid confusion like https://github.com/Tribler/dispersy/pull/160/files#r8156409.

boudewijn-tribler:
retest this please

PR #173, opened 2013-12-05
Allow to pass a candidate list.

whirm:
Also add tests for this feature.

### DIFF
--- a/dispersy.py
+++ b/dispersy.py
@@ -2951,7 +2951,7 @@ ORDER BY global_time""", (meta.database_id, member_database_id)))
                 # CommunityDestination.node_count is allowed to be zero
                 if isinstance(meta.destination, CommunityDestination) and meta.destination.node_count > 0:
                     max_candidates = meta.destination.node_count + len(candidates)
-                    for candidate in islice(meta.community.dispersy_yield_verified_candidates(), max_candidates):
+                    for candidate in meta.community.dispersy_yield_verified_candidates():
                         if len(candidates) < max_candidates:
                             candidates.add(candidate)
                         else:
--- a/dispersy.py
+++ b/dispersy.py
@@ -2942,17 +2942,21 @@ ORDER BY global_time""", (meta.database_id, member_database_id)))
         assert all(message.community == messages[0].community for message in messages)
         assert all(message.meta == messages[0].meta for message in messages)
 
-        result = False
+        result = True
         meta = messages[0].meta
-        if isinstance(meta.destination, CommunityDestination):
-            # CommunityDestination.node_count is allowed to be zero
-            if meta.destination.node_count > 0:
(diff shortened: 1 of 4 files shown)


whirm:
#174 needs to be merged first for this one to work

boudewijn-tribler:
retest this please

dispersy.py
```diff
@@ -2877,17 +2877,17 @@ def _forward(self, messages):
...
-            result = all(self._send(message.destination.candidates, [message]) for message in messages)
-
+        if isinstance(meta.destination, (CommunityDestination, CandidateDestination)):
+            for message in messages:
+                candidates = set()
+                # CommunityDestination.node_count is allowed to be zero
+                if isinstance(meta.destination, CommunityDestination) and meta.destination.node_count > 0:
+                    candidates.update(islice(meta.community.dispersy_yield_verified_candidates(), meta.destination.node_count))
+                # CandidateDestination.candidates may be empty
+                candidates.update(message.destination.candidates)
```
boudewijn-tribler: If we add `CandidateDestination.candidates` to the `set` first, followed by verified candidates until we have either `node_count` or `node_count + len(CandidateDestination.candidates)` candidates to send the message to (or we run out of verified candidates).

This would result in more consistent behaviour because the random choices made in `dispersy_yield_verified_candidates` is not a factor any more.  This will also make it possible to make the unit tests test more deterministic, since we know exactly how many candidates should receive a message.

dispersy.py
```diff
@@ -2877,17 +2877,21 @@ def _forward(self, messages):
...
+                # CandidateDestination.candidates may be empty
+                candidates = set(message.destination.candidates)
+                # CommunityDestination.node_count is allowed to be zero
+                if isinstance(meta.destination, CommunityDestination) and meta.destination.node_count > 0:
+                    max_candidates = meta.destination.node_count + len(candidates)
+                    for candidate in islice(meta.community.dispersy_yield_verified_candidates(), max_candidates):
+                        if len(candidates) < max_candidates:
+                            candidates.add(candidate)
+                        else:
+                            break
```
boudewijn-tribler, line 2958: The `islice` can cause some unexpected behaviour.

For example: when the neighbourhood has `[A,B,C,D]` as verified candidates, `node_count=3`, and `message.destination.candidates=[B,C,D]`.  It is possible that `yield_verified_candidates` returns `[B,C,D]` (since it returns candidates in random order).  This would result the message being sent to only `[B,C,D]` while you intended it to go to `[A,B,C,D]`.

Perhaps this should also be covered in a unit test (if it isn't already).  When you are in the office I can show you a trick to tailor the behaviour of a unit test to 'trigger' the above corner case.

boudewijn-tribler:
Reminder: add deterministic unit test

Issue #167, opened 2013-11-26
Callback.call has a delay argument


NielsZeilemaker:
Callback.call which performs a blocking call has a delay argument.
I don't see a usecase for that, as it will only make you block more.

Shall we remove it?

boudewijn-tribler:
I agree.

PR #175, opened 2013-12-06
Let Callback.call warn on timeout.

boudewijn-tribler:


NielsZeilemaker:
Could you add/remove #167 to this pull request?

PR #176, opened 2013-12-06
Fixes duplicate callback identifier bug.

boudewijn-tribler:
Using RequestCache identifiers (which are unique per community) for
Callback identifiers (which should be unique per Dispersy) will
obviously cause conflicts.

### DIFF
--- a/requestcache.py
+++ b/requestcache.py
@@ -8,16 +8,48 @@ class Cache(object):
 
     @staticmethod
     def create_identifier():
+        """
+        Create an identifier, preferably unique for each outstanding request cache.
+        """
         raise NotImplementedError()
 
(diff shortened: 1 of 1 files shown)


NielsZeilemaker:
Thanks, this look really nice. Are you planning to add the community identifier to all identifiers or is that a different pull request?

NielsZeilemaker:
Nevermind, that isn't neccesary anymore.

boudewijn-tribler:
I wasn't planning on doing that at all tbh, now that the `RequestCache` is unique per community it doesn't require the community id in the identifier string.  Or do you have a use case in mind where this would be useful?

boudewijn-tribler:
Ah ok :)

PR #179, opened 2013-12-10
Infinite loop bug

boudewijn-tribler:
The bug is described and triggerd in the unit test included in this pull request.


boudewijn-tribler:
We can make this simpler.

PR #186, opened 2013-12-11
Simplifies Dispersy.stop.

boudewijn-tribler:
Dispersy.stop calls Callback.stop, which now takes a final_func
parameter.  This is called (once) when Callback.stop has shutdown all
existing tasks.

### DIFF
--- a/dispersy.py
+++ b/dispersy.py
@@ -4585,18 +4585,21 @@ ORDER BY global_time""", (meta.database_id, member_database_id)))
 
         This method is thread safe.
 
-        1. unload all communities
+        1. stops callback
+           a. new tasks are no longer accepted
+           b. flushes existing tasks
+           c. stops existing generators
+        2. unload all communities
            in reverse define_auto_load order, starting with all undefined communities
-        2. closes endpoint
-        3. closes database
-        4. stops callback
+        3. closes endpoint
+        4. closes database
 
-        Returns True when all above steps were successfully completed, otherwise False is returned.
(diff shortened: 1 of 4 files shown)


dispersy.py
```diff
@@ -4626,67 +4629,39 @@ def ordered_unload_communities():
...
+            results[u"callback"] = self._callback.stop(timeout, final_func=stop)
 
-        if self._callback.is_current_thread:
-            # the result from _callback.stop will always be False since it can not stop a thread
-            # that we are currently on
-            self._callback.stop(timeout)
-        else:
-            results.append((u"callback", self._callback.stop(timeout)))
-            assert all(isinstance(result, bool) for _, result in results), [type(result) for _, result in results]
+            if not self._callback.is_current_thread:
```
NielsZeilemaker: Could you explain this line to me? Callback.stop is blocking for timeout seconds right? Why would we like to join the callback again with the same timeout if we're not on the callback thread.

boudewijn-tribler: Some of my unit tests were failing because the stop was called recursively (the test was designed to generate exceptions).  Stop was called once from the callback thread (not allowing the thread to actually close) and once from the nosetests thread, where the stop was handled differently because it had already been closed.  Not waiting for the thread to actually close was causing some issues.

Since then I've added some checks to prevent this recursion which apparently had the side effect of no longer needing the join that you refer to.  I'll commit a change to remove it.

PR #193, opened 2013-12-18
Fixes #136 (RequestCache.pop() behavior)

### LINKED ISSUE #136: RequestCache.pop() behavior
First discussed here https://github.com/Tribler/dispersy/pull/106#discussion_r5369978.

We should improve the `pop` behavior, it doesn't make sense.
- We could remove the cleanup_delay entirely, although this would mean that we no longer have the ability to ensure that an identifier/number isn't used for some time.  It is currently used for the introduction request, adding 4.5 seconds before the identifier can be used again.  Without the cleanup_delay the randomness in the numbers will have to prevent clashes.
- We could add the identifier to a different dictionary or set, removing it from `RequestCache_identifiers`.  This would result in has, get, and pop no longer being able to find the identifier.  We would need a new way to see if an identifier is unclaimed though, for instance by introducing `is_used` or something similar that checks `_identifiers` and the new dict/set.
- I'm sure there are other options.

Anyway, removing the cleanup_delay entirely is the simplest change, and would result in a simplification of the RequestCache.  Are there cases where the delay is essential?

whirm:


whirm:
We are going to test it in some experiment before merging it to devel, so don't merge it yet.

PR #200, opened 2014-01-13
Move initiate meta messages

whirm:


whirm:
This is not yet ready to merge

PR #201, opened 2014-01-15
Do not raise an exception if the candidate is not in self._candidates.

whirm:


NielsZeilemaker:
@whirm we have to look into this, it doesn't make sense that other.sock_addr isn't in self._candidates as the "others" are first extracted from the self._candidate list

whirm:
I can't reproduce it anymore, I'm closing the pull request

NielsZeilemaker:
Do you want to commit the fix for the wrong indentation?

whirm:
oh, true, give me a minute

PR #203, opened 2014-01-16
Move take_step() from Dispersy to Community

whirm:


tests/test_bootstrap.py
```diff
@@ -135,7 +135,7 @@ def dispersy_enable_candidate_walker(self):
             def dispersy_enable_candidate_walker_responses(self):
                 return True
 
-            def dispersy_take_step(self):
+            def take_step(self):
```
NielsZeilemaker: I guess this method needs to accept the allow_sync property

community.py
```diff
@@ -1150,9 +1150,18 @@ def add_conversion(self, conversion):
...
         self._conversions.append(conversion)
 
-    @documentation(Dispersy.take_step)
-    def dispersy_take_step(self, allow_sync):
-        return self._dispersy.take_step(self, allow_sync)
+    def take_step(self, allow_sync):
+        if self.cid in self._dispersy._communities:
+            candidate = self.dispersy_get_walk_candidate()
+            if candidate:
+                assert self.my_member.private_key
```
NielsZeilemaker: Drop this assert, doesn't make sense here

PR #216, opened 2014-01-28
Decode message fix

whirm:


whirm:
Don't merge yet, I have to test it with AllChannel

NielsZeilemaker:
Seems like a great fix, however I think we might have stumbled upon a problem. What if I have seen a member in one community but not in the other. This will cause me to re-use the publickey from the "one" community to verify the packets in the "other" community.

What if a peer requests the missing member in the "other" community? We won't be able to fullfill the request as we only have an identify message in the "one" community. 

Short summary, I think the get_member_from_id should be moved to community.py and the cache etc. Moreover, this requires a change to the member table as it requires a community column.

NielsZeilemaker:
It's not as bad as I expected, however it's implemented in a strange manner. A member object has a has_identity method which has a community as an argument. If this returns false, the identity message is requests.

Still my proposed change can merge these two into a single method call.

whirm:
should we merge this one and go for the next refactor step? (the experiment ran fine)

PR #219, opened 2014-02-04
Add community version when sorting/grouping

whirm:


community.py
```diff
@@ -1859,7 +1859,7 @@ def on_incoming_packets(self, packets, cache=True, timestamp=0.0):
         assert isinstance(timestamp, float), timestamp
 
         messages = []
-        for message_type, iterator in groupby(packets, key=lambda tup: tup[1][22]):
+        for message_type, iterator in groupby(packets, key=lambda tup: (tup[1][1], tup[1][22])):
```
NielsZeilemaker: I think this message_type now is a tuple containing the community_version, message_type right?
I guess you need to unwrap it to get only the message_type

whirm: We aren't using the message type either, Ill replace it with an underscore.

NielsZeilemaker:
retest this please

PR #223, opened 2014-02-05
Small cleanup for candidate

LipuFei:
- Remove unused method get_destination_address().


NielsZeilemaker:
I'll merge it, however this method was used until very recently in Tribler.
Maybe, we should have a policy regarding deprecated methods. I.e. keep them for X months?

PR #234, opened 2014-02-19
Rewrote most unit-tests, still need a lot of cleanup.

NielsZeilemaker:
- Removed legacy code from dispersy
- Removed crazy if **debug** imports


authentication.py
```diff
@@ -37,8 +36,7 @@ def is_signed(self):
             pass
 
         def setup(self, message_impl):
-            if __debug__:
-                from .message import Message
+            from .message import Message
```
whirm, line 39: Is there a reason for not doing the imports all at once at the beginning?

NielsZeilemaker, line 39: It will cause cyclic imports, e.g. dispersy imports community, community imports dispersy

PR #236, opened 2014-03-10
Slightly improve decode ip/port addresses, doesn't create substrings

NielsZeilemaker:
anymore


NielsZeilemaker:
@whirm ready to be merged

PR #244, opened 2014-04-10
Added two assert which verify that walk and stumble candidates have at

NielsZeilemaker:
least one member associated


NielsZeilemaker:
@whirm ready to be merged

PR #245, opened 2014-04-10
Remove 'cluster' column from meta_message table

LipuFei:
- Removed some unused imports.
- Removed 'cluster' column from meta_message table

### DIFF
--- a/dispersydatabase.py
+++ b/dispersydatabase.py
@@ -13,7 +13,7 @@ from .distribution import FullSyncDistribution
 from .logger import get_logger
 logger = get_logger(__name__)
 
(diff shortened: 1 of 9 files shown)


dispersydatabase.py
```diff
@@ -458,6 +456,30 @@ def check_database(self, database_version):
 
             new_db_version = 21
             if database_version < new_db_version:
+                # remove 'cluster' column from meta_message table
+                logger.debug("upgrade database %d -> %d", database_version, 18)
```
NielsZeilemaker: could you change the logger to use new_db_version instead of 18

PR #253, opened 2014-04-15
Fix for sequence number check, messages aren't inserted in order.

NielsZeilemaker:
Hence, we can only check after inserting all messages.


NielsZeilemaker:
retest this please

PR #273, opened 2014-05-08
small cleanup

LipuFei:


LipuFei:
@NielsZeilemaker @whirm ready

dispersy.py
```diff
@@ -2196,15 +2165,6 @@ def _flush_database(self):
             # OperationalError: database is locked
             logger.exception("%s", exception)
 
-    # TODO this -private- method is not used by Dispersy (only from the Tribler SearchGridManager).
```
NielsZeilemaker, line 2199: Could you submit a pullrequest for tribler changing this?

endpoint.py
```diff
@@ -78,10 +80,21 @@ def close(self, timeout=0.0):
...
     def log_packet(self, sock_addr, packet, outbound=True):
+        name = "???"
         try:
-            name = self._dispersy.convert_packet_to_meta_message(packet, load=False, auto_load=False).name
+            community = self._dispersy.get_community(packet[2:22], load=False, auto_load=False)
+
+            # find associated conversion
+            conversion = community.get_conversion_for_packet(packet)
+            name = conversion.decode_meta_message(packet).name
+        except CommunityNotFoundException:
```
NielsZeilemaker: You can remove these except clauses, only keeping the expect: pass

dispersy.py
LipuFei, line 2199: I did a search in tribler and actually there is nothing that uses this method, so I guess there is no need to change tribler.

LipuFei, line 2199: Oh sorry... SearchGridManager.py actually uses it. I'll make a pull request for this.

PR #275, opened 2014-05-08
Fix the create identity message checking

whirm:


whirm:
@NielsZeilemaker Does this look OK to you?

PR #289, opened 2014-05-21
Using the WalkCandidate stored in the request

NielsZeilemaker:
Walkcandidates then to get lost in Dispersy. Hence, we need to use the one when we attempted to
send the introduction_request.


NielsZeilemaker:
@egbertbouman @whirm ready to be merged

PR #292, opened 2014-05-22
Improved sorting in Discovery Community

NielsZeilemaker:
- sorting tbs by random if they have the same overlap
- sorting possible by timestamp if they have the same overlap, then by random
- limiting to max_tbs to maintain connections to


NielsZeilemaker:
retest this please

PR #300, opened 2014-05-26
Endpoint migration to twisted

whirm:
WIP: DO NOT MERGE YET

### DIFF
--- a/endpoint.py
+++ b/endpoint.py
@@ -5,10 +5,11 @@ import sys
 import threading
 from abc import ABCMeta, abstractmethod
 from itertools import product
-from select import select
 from time import time
 
 from twisted.internet import reactor
+from twisted.internet.defer import Deferred, gatherResults, succeed
+from twisted.internet.protocol import DatagramProtocol
 
 from .candidate import Candidate
 from .logger import get_logger
@@ -22,6 +23,15 @@ else:
     SOCKET_BLOCK_ERRORCODE = errno.EWOULDBLOCK
 
 TUNNEL_PREFIX = "ffffffff".decode("HEX")
+TUNNEL_PREFIX_LENGHT = 4
+
+def strip_if_tunnel(datagram):
+    """
+    Returns is_tunnel, prefix_stripped_datagram
+    """
+    if datagram.startswith(TUNNEL_PREFIX):
+        return True, datagram[TUNNEL_PREFIX_LENGHT:]
+    return False, datagram
 
 
 class Endpoint(object):
@@ -60,7 +70,8 @@ class Endpoint(object):
             name = conversion.decode_meta_message(packet).name
         except:
             name = "???"
-        logger.debug("%30s %s %15s:%-5d %4d bytes", name, '->'if outbound else '<-', sock_addr[0], sock_addr[1], len(packet))
+        logger.debug("%30s %s %15s:%-5d %4d bytes", name, '->' if outbound else '<-',
+                     sock_addr[0], sock_addr[1], len(packet))
 
         if outbound:
             self._dispersy.statistics.dict_inc(u"endpoint_send", name)
@@ -235,8 +246,7 @@ class RawserverEndpoint(Endpoint):
 
                 self._dispersy.statistics.cur_sendqueue = len(self._sendqueue)
 
-
-class StandaloneEndpoint(RawserverEndpoint):
+class StandaloneEndpoint(RawserverEndpoint, DatagramProtocol):
 
     def __init__(self, port, ip="0.0.0.0"):
         # do NOT call RawserverEndpoint.__init__!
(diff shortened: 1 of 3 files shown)


endpoint.py
```diff
@@ -58,9 +68,11 @@ def log_packet(self, sock_addr, packet, outbound=True):
             # find associated conversion
             conversion = community.get_conversion_for_packet(packet)
             name = conversion.decode_meta_message(packet).name
+        # TODO(emilon): Shouldn't we be a bit more selective here?
```
NielsZeilemaker: No, if anything goes wrong we'll just use ??? as a name.

endpoint.py
```diff
@@ -129,7 +141,8 @@ def close(self, timeout=0.0):
 
     def get_address(self):
         assert self._dispersy, "Should not be called before open(...)"
-        return self._socket.getsockname()
+        address = self.transport.getHost()
```
NielsZeilemaker: RawserverEndpoint does not have a self.transport, StandAloneEndpoint does

endpoint.py
```diff
@@ -245,125 +257,99 @@ def __init__(self, port, ip="0.0.0.0"):
...
         result = True
 
-        if timeout > 0.0:
-            self._thread.join(timeout)
+        self.transport.stopListening()
 
-            if self._thread.is_alive():
-                logger.error("the endpoint thread is still running (after waiting %f seconds)", timeout)
-                result = False
+        # TODO(emilon): Maybe move this stuff to an endpoint subclass? I think it's only going to be use by the tests.
```
NielsZeilemaker: This has to be moved to the endpoint class, as its used in dispersy.stop

endpoint.py
```diff
@@ -245,125 +257,99 @@ def __init__(self, port, ip="0.0.0.0"):
...
 
 class ManualEnpoint(StandaloneEndpoint):
 
     def __init__(self, *args, **kwargs):
         StandaloneEndpoint.__init__(self, *args, **kwargs)
-        self.receive_lock = threading.RLock()
         self.received_packets = []
 
+    # TODO(emilon): deprecate this?
     def data_came_in(self, packets):
```
NielsZeilemaker: Just remove it

endpoint.py
```diff
@@ -245,125 +257,99 @@ def __init__(self, port, ip="0.0.0.0"):
...
+                     len(packets), len(self.received_packets))
 
-        with self.receive_lock:
-            self.received_packets.extend(packets)
+    def datagramReceived(self, datagram, address):
+        logger.debug('added packet to receivequeue, %d packets are queued in total', len(self.received_packets))
+        self.received_packets.append((address, datagram))
 
     def clear_receive_queue(self):
-        with self.receive_lock:
```
NielsZeilemaker, line 361: I'm not sure which threads are going to be calling these methods, hence i'd rather keep the lock

endpoint.py
whirm: the endpoint class is not derivating from DatagramProtocol, This is Only used on the tests so maybe we could move it to ManualEndpoint instead

endpoint.py
whirm, line 361: I would just decorate the method

endpoint.py
NielsZeilemaker: Look at dispersy.py:2214

```
# Wait for the endpoint's port to be closed
results[u"endpoint disconnect"] = maybeDeferred(self._endpoint.disconnection_lock.acquire)
```

endpoint.py
NielsZeilemaker, line 361: Just add the lock, it won't cause any overhead.

endpoint.py
whirm: oh, right...

NielsZeilemaker: Maybe a nicer solution would be to return this disconnection_lock and let dispersy wait for that deferred, instead of forcing all other endpoint to "call" this deferred.

whirm: but the endpoint is an abstract class, should we move all the datagramprotocol stuff there and replace the standalone endpoint with it? 
Can you join #tribler?

endpoint.py
```diff
@@ -245,104 +255,75 @@ def __init__(self, port, ip="0.0.0.0"):
...
-                logger.error("the endpoint thread is still running (after waiting %f seconds)", timeout)
-                result = False
+        self.transport.d.addBoth(on_closed)
 
-        else:
-            if self._thread.is_alive():
-                logger.debug("the endpoint thread is still running (use timeout > 0.0 to ensure the thread stops)")
-                result = False
+        # do NOT call RawserverEndpoint.close!
+        return gatherResults((d, maybeDeferred(Endpoint.close, self, timeout)), consumeErrors=True).addCallback(all)
```
NielsZeilemaker: Don't assume that endpoint close may return a deferred. Just call it directly.

PR #304, opened 2014-05-27
Fix test overlay

whirm:


NielsZeilemaker:
@whirm while you're at it, can you remove the bartercast one.

whirm:
I would say it's ready to be merged

PR #306, opened 2014-05-28
Peer cache for DiscoveryCommunity

egbertbouman:


discovery/community.py
```diff
@@ -185,6 +188,10 @@ def insert_trackers(self):
                     logger.debug("Adding %s %s as discovered candidate", type(community), candidate)
                     community.add_discovered_candidate(candidate)
 
+    def dispersy_get_walk_candidate(self):
+        candidate = super(DiscoveryCommunity, self).dispersy_get_walk_candidate()
+        return candidate or self.peer_cache.get_peer()
```
NielsZeilemaker, line 193: Won't this get_peer method now get called every time? Maybe an if is a better solution.

discovery/community.py
```diff
@@ -306,6 +315,8 @@ def reset_taste_buddy(self, candidate):
         for tb in self.yield_taste_buddies():
             if tb == candidate:
                 tb.timestamp = time()
+                if candidate.connection_type == u'public':
```
NielsZeilemaker: You could use tb.should_cache() instead

discovery/community.py
```diff
@@ -632,7 +645,7 @@ def on_timeout(self):
                     self.community.remove_taste_buddy(candidate)
 
     def create_ping_requests(self):
-        tbs = [tb.candidate for tb in self.yield_taste_buddies() if tb.time_remaining() < PING_INTERVAL]
+        tbs = [tb.candidate for tb in self.yield_taste_buddies() if tb.time_remaining() < PING_INTERVAL][:self.max_tbs]
```
NielsZeilemaker: I guess this needs to be changed to

```
tbs = list(self.yield_taste_buddies())[:self.max_tbs]
tbs = [tb.candidate for tb in tbs if tb.time_remaining() < PING_INTERVAL]
```

discovery/community.py
```diff
@@ -678,3 +691,88 @@ def _create_pingpong(self, meta_name, candidates, identifier):
...
+
+        if len(self.walkcandidates) > self.walkcandidates_limit:
+            sorted_keys = sorted([(info['last_seen'], wcandidate) for wcandidate, info in self.walkcandidates.iteritems()], reverse=True)
+            for _, wcandidate in sorted_keys[:self.walkcandidates_limit]:
+                del self.walkcandidates[wcandidate]
+
+        logger.debug('PeerCache: removed %d peers', old_num_candidates - len(self.walkcandidates))
+
+        with open(self.filename, 'w') as fp:
+            lines = ['# WAN address\t\tLAN address\t\tTunnel\tLast seen\t\tLast checked\tNumber of fails\n']
```
NielsZeilemaker: I would convert these into prints, i.e.

```
print >> fp, '# WAN address\tLAN address\tTunnel etc.
```

discovery/community.py
```diff
@@ -678,3 +691,88 @@ def _create_pingpong(self, meta_name, candidates, identifier):
...
+            for wcandidate, info in self.walkcandidates.iteritems():
+                line = '%s:%d\t%s:%d\t%r\t' % (wcandidate.wan_address + wcandidate.lan_address + (wcandidate.tunnel,))
```
NielsZeilemaker: Same here

discovery/community.py
egbertbouman, line 193: I don't think so, self.peer_cache.get_peer() is only executed is candidate is None.

NielsZeilemaker, line 193: I tried it, you're right it doesn't call the method.

PR #338, opened 2014-06-13
start and stop should always be called from the reactor thread

whirm:


whirm:
When this one gets merged Ill send a pull request to Tribler with the necessary stuff and the ref. update

whirm:
@NielsZeilemaker , @LipuFei ready

PR #340, opened 2014-06-25
Improve blockingCallFromThread so it shows the full stacktrace from both...

whirm:
... threads

when an exception is raised.

### DIFF
--- a/util.py
+++ b/util.py
@@ -1,5 +1,7 @@
+import Queue
 import functools
 import sys
+import traceback
 import warnings
 from cProfile import Profile
 from collections import defaultdict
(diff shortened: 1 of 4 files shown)


whirm:
@LipuFei, @NielsZeilemaker ready to merge if the tests pass

util.py
```diff
@@ -9,11 +10,18 @@
...
 from twisted.internet import reactor
 from twisted.internet.task import LoopingCall
-from twisted.internet.threads import blockingCallFromThread
 from twisted.python.threadable import isInIOThread
+from twisted.internet import defer
+from twisted.python import failure
 
 from .logger import get_logger
 
+from twisted.python.compat import _PY3
```
NielsZeilemaker: I don't think the rest of Dispersy is compatible with Python3, why add this here?

NielsZeilemaker:
Except for the compatibility thingy, the rest seems fine by me.

PR #345, opened 2014-07-23
Fix test candidates

whirm:
This PR fixes test_candidates.py, uncovering a bug we have in the walk candidates iterator, so it will fail until it gets fixed.


whirm:
Ready to merge if the test passes

PR #348, opened 2014-07-24
Per-class logger

LipuFei:
Use per-class logger.


LipuFei:
@whirm I think this should be it. I've rearranged the order of the commits, so e69fb04 is actually the last commit.

whirm:
retest this please

whirm:
@LipuFei some lines got over 120 chars with your updates, could you fix that? 
I noticed pylint wasn't finding its config file, I'm fixing that now.

whirm:
Other than that looks OK to me

LipuFei:
@whirm I've solved most of those pep8 problems.

PR #350, opened 2014-07-25
Initialize a community before adding it to dispersy

LipuFei:
Initialize first otherwise some threading issues may occur.


whirm:
This would undo the memory leak fix we merged yesterday.

PR #354, opened 2014-07-31
Fix encoding check for V2 community

LipuFei:


LipuFei:
@whirm ready to be merged

PR #355, opened 2014-08-06
Initialize some variables in the constructor

whirm:


whirm:
@LipuFei ready to merge, works ok on allchannel and locally with Tribler

PR #356, opened 2014-08-08
Start walking after adding the bootstrap candidates.

whirm:
Not before :)


whirm:
@LipuFei ready to merge

PR #359, opened 2014-08-14
If all the Discovery community's candidate sources are empty for a minut...

whirm:
...e, use a randomly selected tracker to take the step.


whirm:
@LipuFei ready

PR #367, opened 2014-08-28
Go back to using a threaded endpoint insted of a twisted protocol.

whirm:
To work around the fact that the database blocks for so long that the UDP buffers fill and loads
of messages get lost.


whirm:
Ready to merge

PR #370, opened 2014-09-09
WIP: Cancel message

LipuFei:
Use cancel message to replace undo message.
All undo message code remain the same.

### DIFF
--- a/community.py
+++ b/community.py
@@ -3406,15 +3406,88 @@ class Community(TaskManager):
         """
         assert all(message.name in (u"dispersy-cancel-own", u"dispersy-cancel-other") for message in messages)
 
-        # We first need to extract the DispersyDuplicatedUndo objects from the messages list and deal with them
-        real_messages = list()
-        parameters = list()
+        # Check the cancel messages we have got and only update the database with the latest one
+        # (which is the message with the biggest mid and global_time)
+        # check the cancel messages grouped by the message they cancel
+        categoried_msg_dict = dict()
         for message in messages:
             if isinstance(message, Message.Implementation) and message.payload.process_cancel:
                 # That's a normal cancel message
-                parameters.append((message.packet_id, self.database_id, message.payload.member.database_id,
-                                   message.payload.global_time))
-                real_messages.append(message)
+                #parameters.append((message.packet_id, self.database_id, message.payload.member.database_id,
+                #                   message.payload.global_time))
+                key = (message.payload.global_time, message.payload.member.mid)
+                if key not in categoried_msg_dict:
+                    categoried_msg_dict[key] = dict()
+                    categoried_msg_dict[key]["messages"] = list()
+
+                    # try to load the cancel message we have in the database
+                    undone, = self._dispersy._database.execute(
+                        u"SELECT undone FROM sync WHERE community = ? AND member = ? AND global_time = ?",
+                        (self.database_id, message.payload.member.database_id, message.payload.global_time)).next()
+                    if undone != 0:
+                        try:
+                            mid, global_time = self._dispersy._database.execute(
+                                u"SELECT member.mid, sync.global_time FROM sync"
+                                u" JOIN meta_message ON meta_message.id == sync.meta_message"
+                                u" JOIN member ON member.id == sync.member"
+                                u" WHERE meta_message.name == 'CancelMessage' AND sync.id == ?", (undone,)).next()
+                        except StopIteration:
+                            # no previous cancel for this message
+                            pass
+                        else:
+                            categoried_msg_dict[key]["cancel_in_db"] = (global_time, mid, undone)
+
+                categoried_msg_dict[key]["messages"].append(message)
+
+        # for each message to be canceled, check the cancel messages member global_time and mid
+        # (the larger the later), and we only use the latest one.
+        # The logic is that:
+        # (1) if the latest message is already in our database:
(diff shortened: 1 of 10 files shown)


LipuFei:
@whirm I'll remove the pep8 commit for now, otherwise the PR is too difficult to review.

whirm:
@LipuFei yes, those should always go alone

NielsZeilemaker:
@LipuFei why are you adding the cancel to the permission triplet? That shouldn't be neccesary. You authorise someone to send a dispersy_cancel message etc.

LipuFei:
@NielsZeilemaker I made this PR to make it easy for us to review. Thanks! I'll check everything again.

LipuFei:
@NielsZeilemaker What does the `permission_triplets` mean?

NielsZeilemaker:
Basically, you can authorize/revoke or grant someone the permission to create a message of a certain type. I wouldn't start messing with this as its never tested/understood by me.
Moreover, are you sure you want to introduce a cancel callback? Maybe calling the undo callback would suffice. Chaning it slightly to allow you to pass a cancelled/undone/done variable instead of the undone/done currently used.

NielsZeilemaker:
Btw, the documentation regarding the permissions should still be correct. You can have a look there.

LipuFei:
@NielsZeilemaker I had thought about it and I decided to duplicate the undo stuff instead of messing them up. Later, when we want to remove the undo, we can simply remove every undo-related thing without worry about causing damage to the cancel message. I will read the permission docs. Thanks!

community.py
```diff
@@ -132,11 +132,12 @@ def create_community(cls, dispersy, my_member, *args, **kargs):
...
         # authorize MY_MEMBER
         permission_triplets = []
-        message_names = (u"dispersy-authorize", u"dispersy-revoke", u"dispersy-undo-own", u"dispersy-undo-other")
+        message_names = (u"dispersy-authorize", u"dispersy-revoke", u"dispersy-undo-own", u"dispersy-undo-other",
+                         u"dispersy-cancel-own", u"dispersy-cancel-other")
         for message in community.get_meta_messages():
             # grant all permissions for messages that use LinearResolution or DynamicResolution
             if isinstance(message.resolution, (LinearResolution, DynamicResolution)):
-                for allowed in (u"authorize", u"revoke", u"permit"):
+                for allowed in (u"authorize", u"revoke", u"permit", u"cancel"):
```
NielsZeilemaker: I would remove the cancel here, but leave it in timeline.py. E.g. add it when the "undo" thingy is there.

community.py
```diff
@@ -3114,7 +3144,7 @@ def create_authorize(self, permission_triplets, sign_with_master=False, store=Tr
                 assert isinstance(triplet[0], Member)
                 assert isinstance(triplet[1], Message)
                 assert isinstance(triplet[2], unicode)
-                assert triplet[2] in (u"permit", u"authorize", u"revoke", u"undo")
+                assert triplet[2] in (u"permit", u"authorize", u"revoke", u"undo", u"cancel")
```
NielsZeilemaker, line 3207: So here is fine I guess :smile:

tests/test_cancel.py
```diff
@@ -0,0 +1,217 @@
...
+
+        # create messages
+        messages = [node.create_full_sync_text("Should cancel #%d" % i, i + 10) for i in xrange(10)]
+        node.give_messages(messages, node)
+
+        # check that they are in the database and are NOT undone
+        node.assert_is_stored(messages=messages)
+
+        # undo all messages
+        cancels = [node.create_undo_own(message, i + 100) for i, message in enumerate(messages)]
```
NielsZeilemaker: You're calling create_undo here instead of create_cancel

LipuFei:
@whirm I guess this is more or less it. I will check the unit tests one more time after my presentation rehearsal. Also, before we merge, we should change the AllChannel community to use this cancel message and test on jenkins.

whirm:
Ok, let's review it later

LipuFei:
@NielsZeilemaker One of the cancel message tests didn't pass. The test does the following:

```
Three nodes: SELF, NODE, OTHER
What they do:
1. SELF gives NODE the permission to cancel.
2. OTHER created message M1.
3. NODE cancels M1 with cancel message C1.
4. M1 should be found as cancelled by NODE (This one passed).
5. SELF revokes NODE's cancel permission.
6. M1 should still be marked as cancelled by NODE (This one failed).
```

This test was passing before. Am I doing it correctly? If a user's cancel permission has been revoked, should his old cancel messages be regarded as invalid as well?

NielsZeilemaker:
@LipuFei maybe create a wiki page describing the cancel message, and what it should do. Then afterwards, I can have a look and maybe we can modify the behaviour of your implementation accordingly.
Also describe situations such as the one above. A gets the permission to cancel other, A's permission gets revoke -> what should be do with the messages cancelled by A.

LipuFei:
@NielsZeilemaker I've made a wiki page: https://github.com/Tribler/dispersy/wiki/Cancel-Message. Hope it explains.

NielsZeilemaker:
@LipuFei I made some changes to the document. I guess cancel messages should also not cancel an authorize message right? Or are we adding that? Additionally, we should think about being able to drop cancelled messages from the database in order to save space.

LipuFei:
Hi @NielsZeilemaker, could you review this PR again? Thanks!

PR #373, opened 2014-10-08
Fix a problem that file paths can be unicode

### LINKED ISSUE #871: Unicode charaters in path crash Tribler for Windows
Hi!

I downloaded Tribler a couple days ago (pre-compiled installer), but couldn't get it to start. It always showed an error window for like half a second and then died. Perhaps that's another bug? On fatal crash it just dies without allowing the user to read the error.

Anyway, so I have not been able to test it yet. I tried installing it a couple times, and installing a previous version. I checked the log file until today (I know I should've done this first), and found this (skipping some debug lines to prove a point):

```
Traceback (most recent call last):
...
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe9 in position 14: ordinal not in range(128)
...
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe9 in position 14: ordinal not in range(128)
ERROR   1410953058.90   sqlitecachedb:369   No existing database found. Attempting to create a new database u'C:\\Users\\Sinuh\xe9\\AppData\\Roaming\\.Tribler\\sqlite\\tribler.sdb'
ERROR   1410953060.24            util:269   Exception raised on the reactor's thread.
...
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 14: ordinal not in range(128)
ERROR   1410953560.14            util:269   Exception raised on the reactor's thread.
 Traceback from this thread:
...
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 14: ordinal not in range(128)
```

So the problem is, I think very obviously, that Tribler is using ASCII encoder for reasons I don't understand (yet, of course). Windows uses UTF-16 for its filename system, and Linux + Mac use UTF-8 (as far as I know)... and that's happening since the early 2000, and since Tribler's using quite modern technologies I see no reason to be _so much_ backwards compatible. And even then, the guys behind UTF-8 and UTF-16 where clever enough to build them to be compatible with ASCII characters, so `"this string"` would be exactly the same as `"this string"` in either encoding...

I'm not a Python kind of guy, so I'm not completely sure as to where should I go about trying to fix this. I'll trust you guys on this one :stuck_out_tongue_winking_eye:

**P.S.:** I've posted this on the [_Tribler's_ forum](http://forum.tribler.org/viewtopic.php?f=6&t=6686) too... I don't know if it's right or wrong.

LipuFei:
Fixes https://github.com/Tribler/tribler/issues/871

### DIFF
--- a/database.py
+++ b/database.py
@@ -248,7 +248,7 @@ class Database(object):
             return False
 
     @attach_explain_query_plan
-    @attach_runtime_statistics("{0.__class__.__name__}.{function_name} {1} [{0.file_path}]")
+    @attach_runtime_statistics(u"{0.__class__.__name__}.{function_name} {1} [{0.file_path}]")
     def execute(self, statement, bindings=(), get_lastrowid=False):
         """
         Execute one SQL statement.
@@ -295,7 +295,7 @@ class Database(object):
             result = self._cursor.lastrowid
         return result
 
-    @attach_runtime_statistics("{0.__class__.__name__}.{function_name} {1} [{0.file_path}]")
+    @attach_runtime_statistics(u"{0.__class__.__name__}.{function_name} {1} [{0.file_path}]")
     def executescript(self, statements):
         assert self._cursor is not None, "Database.close() has been called or Database.open() has not been called"
         assert self._connection is not None, "Database.close() has been called or Database.open() has not been called"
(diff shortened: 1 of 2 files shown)


LipuFei:
retest this please

LipuFei:
retest this please

LipuFei:
@whirm @egbertbouman ready

PR #379, opened 2014-11-05
Modify RawEndpoint

LipuFei:
Modify RawEndpoint to allow sending data to a specific address.


endpoint.py
```diff
@@ -201,6 +201,36 @@ def send(self, candidates, packets, prefix=None):
 
         return send_packet
 
+    def send_packet_to_address(self, address, packet, prefix=None):
```
egbertbouman, line 204: Since you're creating a new function for this, maybe you can also call send_packet_to_address from send_packet (to avoid duplicate code)?

LipuFei:
@egbertbouman Done.

LipuFei:
@egbertbouman Can't merge them for now. Because for TFTP we don't need to append `TUNNEL_PREFIX`.

PR #391, opened 2015-01-14
Add thread assertion to try to catch a rare exception.

whirm:


whirm:
@LipuFei ready to merge.

PR #394, opened 2015-01-15
Ignore unexpected messages

whirm:
Also fix the test_requestcache.py to actually fail if the tests don't pass...


whirm:
@LipuFei ready to merge

PR #397, opened 2015-01-17
use util.is_valid_address in community.py too

whirm:


whirm:
@LipuFei, @NielsZeilemaker ready

PR #402, opened 2015-01-28
Add a DailyFileLog observer for tracker

LipuFei:
Add a DailyFileLog observer for tracker

### DIFF
--- a/scripts/start_tracker.sh
+++ b/scripts/start_tracker.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+_script=$(readlink -f ${BASH_SOURCE[0]})
+BASE_DIR="$(dirname $_script)/../.."
+
+PORT=$1
+RESTART=$(echo "$2" | tr '[:upper:]' '[:lower:]')
+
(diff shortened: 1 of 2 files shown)


tracker/logger.py
```diff
@@ -0,0 +1,20 @@
+"""
+This file contains the DailyLogFile observer that is used by twistd to save
+tracker logs.
+"""
+from twisted.python.log import FileLogObserver
+from twisted.python.logfile import DailyLogFile
+
+# TODO: Hard code the paths for now. Need to find a better solution.
```
whirm: @LipuFei you can add an argument to the tracker plugin for that

whirm: also: https://twistedmatrix.com/documents/14.0.0/core/howto/application.html#customizing-twistd-logging

I would set it like this, we can't use the default one as it would break the statistics stuff, so it's not a good idea to leave it as default.

LipuFei: @whirm the `tracker_plugin.py` is a `MultiService`. How can I change it to an `Application`?

whirm: We don't need to: https://gist.github.com/djfroofy/1394391

PR #403, opened 2015-02-01
Add source on incoming

NielsZeilemaker:
Used in the hidden_tunnel to detect if a message actually originated from a circuit.


NielsZeilemaker:
@whirm ready, this pull request needs to be merged before the new hidden tunnel refactor

PR #407, opened 2015-02-15
Removed the initial_eligable_peers from the fast walker. 

NielsZeilemaker:
Trackers will keep replying to the introduction_requests and hence will stay in the
eligable list. No need for another list.
Added a random float to make candidates eligable sooner, this will help
load balance bootstrapping with the fast walker.


NielsZeilemaker:
@whirm ready

PR #408, opened 2015-02-16
Implemented encode_decode

NielsZeilemaker:
Also needed to fix the permission system, as revoking permissions did not result in any action.
Luckily, revoking permissions was never used in the "wild".


NielsZeilemaker:
@whirm ready

PR #410, opened 2015-03-13
Improve discovery

NielsZeilemaker:


NielsZeilemaker:
@whirm it's ready, but its a bit worrying that the test_bootstrap failed twice. Any clue if we can get it to show some output? The output of the tracker?

whirm:
@NielsZeilemaker I will check later

NielsZeilemaker:
I've discovered a small bug in the bootstrap unit-test, let me fix that first. Then this pull request should be ready to merge.

NielsZeilemaker:
@whirm I've added some logging which hopefully will help us determine why some of the introduction requests are malformed in the discovery community.
It's ready to be merged.

PR #432, opened 2015-06-18
Fixed/extended doublememberauth

NielsZeilemaker:
When a message is modified, the sender signature is reset. But if accepted unmodified, then both signatures would be set after Bob accepts it.

### DIFF
--- a/community.py
+++ b/community.py
@@ -10,10 +10,10 @@ Community instance.
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict, OrderedDict
 from itertools import islice, groupby
+import logging
 from math import ceil
 from random import random, Random, randint, shuffle, uniform
 from time import time
-import logging
 
 from twisted.internet import reactor
 from twisted.internet.defer import inlineCallbacks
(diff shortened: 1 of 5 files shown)


NielsZeilemaker:
@whirm ready

synctext:
Morning, Thnx for refactor!

Can the outgoing signature-request contain a signed msg, extended by receiver with more info and double signed upon return signature-response..

NielsZeilemaker:
The outgoing message contains a signature, but it's reset after Bob modifies the message. Alice will sign it again after accepting the modifications of Bob.

NielsZeilemaker:
@whirm I'm thinking to not call the callback from here: https://github.com/Tribler/dispersy/blob/devel/community.py#L2431
if the message did not change, as by then we already have a valid double signed message.

whirm:
@NielsZeilemaker sound good to me.

NielsZeilemaker:
Ok, I'll create another pull request for that one. Let's merge this one to let @snorberhuis continue testing his community.

PR #443, opened 2015-07-09
Drop duplicated messages on check_pong()

whirm:
Do the same for check_similarity_response()


whirm:
retest this please

whirm:
@NielsZeilemaker ready

NielsZeilemaker:
Unfortunately, this will break the pings as the ping messages are send to more than one peer. 
I'll change that

PR #446, opened 2015-07-20
Implemented a split_function which allow you to append to a double

NielsZeilemaker:
member auth message.


NielsZeilemaker:
@snorberhuis have a look at test_append_response_from_node in test_double_signature.py
 and the split_double_payload method in debugcommunity/community.py
This latter method needs to split the payload into two parts. One part is signed by Alice, the other by Bob.
In this case, I implemented a unit-test wherein Alice signs the message until the "....Allow=Modified" part, and Bob signs the complete message (after appending MODIFIED to the payload).

snorberhuis:
Please see NielsZeilemaker/dispersy#2 it has a fix for a bug in this PR. You probably would want that bug fix. It didn't know a different way of getting that fix into this PR.

snorberhuis:
Please wait for NielsZeilemaker/dispersy#3. This adds support to also allow append using payload with properties.

snorberhuis:
What is blocking this PR?

NielsZeilemaker:
It's ready, let's have @whirm merge it if he agrees.

whirm:
retest this please

PR #449, opened 2015-08-10
Catch an OSError from netifaces

### LINKED ISSUE #448: netiface error
Check https://github.com/Tribler/tribler/issues/1575

Looks like `netifaces.interfaces()` caused the problem. I checked netifaces code and found that an underlying function may raise an `OSError` if `GetAdaptersAddresses()` fails. So I guess we need to catch `OSError` on `netifaces.interfaces()`.

LipuFei:
Closes #448

### DIFF
--- a/dispersy.py
+++ b/dispersy.py
@@ -191,23 +191,26 @@ class Dispersy(TaskManager):
             def __repr__(self):
                 return "<{self.__class__.__name__} \"{self.name}\" addr:{self.address} mask:{self.netmask}>".format(self=self)
 
-        for interface in netifaces.interfaces():
-            try:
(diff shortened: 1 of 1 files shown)


LipuFei:
@whirm ready

whirm:
This should go against `next`

LipuFei:
Ok, I thought we are still on `devel` with dispersy

LipuFei:
@whirm Wait, I just checked. It should be `devel`. the latest commits in `next` are all from 2 years ago.

whirm:
yeah, sorry, I don't know what I was thinking :D

PR #450, opened 2015-08-19
Fix a logger problem

LipuFei:


LipuFei:
@whirm ready when it passes

PR #451, opened 2015-08-24
Fix an encoding issue in endpoint

LipuFei:
Fix an encoding issue in endpoint reported in https://github.com/Tribler/tribler/issues/1602


LipuFei:
@whirm ready

PR #455, opened 2015-09-09
Added DoubleMemberAuthentication.signatures

snorberhuis:
You need the signatures when you respond to a request with properties that you need to set in append mode. The properties in  a payload do not allow you to change them. So you need to create a new message, but you do not want to lose the signatures already set.

This allows you to get the signatures and add them to the new message.

### DIFF
--- a/authentication.py
+++ b/authentication.py
@@ -291,6 +291,14 @@ class DoubleMemberAuthentication(Authentication):
             return self._members
 
         @property
+        def signatures(self):
+            """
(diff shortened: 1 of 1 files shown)


whirm:
ok to test

snorberhuis:
Please merge this, it is blocking Tribler/tribler#1429.

synctext:
@whirm  Can you have a look at this and merge it?

whirm:
ok

PR #456, opened 2015-11-02
Dispersy test expansion to support community testing better

snorberhuis:
These changes allow communities to be better tested.

This is used to test the MultiChain community:  Tribler/tribler#1429.

### DIFF
--- a/tests/dispersytestclass.py
+++ b/tests/dispersytestclass.py
@@ -1,10 +1,17 @@
 import os
 import logging
+import sys
 from unittest import TestCase
+from tempfile import mkdtemp
 
 # Do not (re)move the reactor import, even if we aren't using it
-# (nose starts the reactor in a separate thread when importing this)
-from nose.twistedtools import reactor
+# (nose starts the reactor in a separate thread when importing this)\
+if "twisted.internet.reactor" in sys.modules.keys():
+    """ Tribler already imported the reactor. """
+    from twisted.internet import reactor
+else:
+    from nose.twistedtools import reactor
+
 from twisted.internet.defer import inlineCallbacks, returnValue
 
 from ..discovery.community import PEERCACHE_FILENAME
@@ -18,7 +25,7 @@ from .debugcommunity.node import DebugNode
 # use logger.conf if it exists
 if os.path.exists("logger.conf"):
     # will raise an exception when logger.conf is malformed
-    logging.config.fileConfig("logger.conf")
+    logging.basicConfig(filename="logger.conf")
 # fallback to basic configuration when needed
 logging.basicConfig(format="%(asctime)-15s [%(levelname)s] %(message)s")
 
@@ -38,6 +45,7 @@ class DispersyTestFunc(TestCase):
         self.dispersy_objects = []
 
         self.assertFalse(reactor.getDelayedCalls())
+        """" Central node that is also used for master member. """
         self._mm = None
         self._mm, = self.create_nodes()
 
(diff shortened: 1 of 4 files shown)


snorberhuis:
I will remove the WIP commit.

whirm:
Ok, so that's what's happening: The delayed call gets canceled from the taskmanager, but the delayed call that calls the start method: 

`reactor.callLater(0, lc.start, DOWNLOAD_MM_PK_INTERVAL, now=True)` in community:346

Hasn't been called yet when `on_identity()` gets called, so the looping call gets started even after being canceled from the taskManager.

snorberhuis:
So this is a separate issue with Dispersy that is a potential bug? 

If I should fix it here, what is the solution? Because there is probably a reason the `reactor.callLater` is used. Can the `callLater` also be removed when canceling the looping call?

whirm:
TaskManager should be extended to create and deal with delayed looping calls so it knows it has to cancel the associated DelayedCall, I'll do it after I fix the Tribler bug I'm working on now. Should be done later today.

whirm:
@snorberhuis: #457 should fix it.

whirm:
@snorberhuis merged

snorberhuis:
I tested with #457 and it does work. I reverted the old bugfix.

tests/dispersytestclass.py
```diff
@@ -18,7 +25,7 @@
 # use logger.conf if it exists
 if os.path.exists("logger.conf"):
     # will raise an exception when logger.conf is malformed
-    logging.config.fileConfig("logger.conf")
+    logging.basicConfig(filename="logger.conf")
```
whirm, line 28: @snorberhuis what issue where you having with this?

Also, better get used to split commits into different features, not a bunch of unrelated stuff.

tests/dispersytestclass.py
```diff
@@ -63,24 +71,43 @@ def tearDown(self):
...
 
                 nodes.append(node)
             self._logger.debug("create_nodes, nodes created: %s", nodes)
             returnValue(nodes)
 
-        return blockingCallFromThread(reactor, _create_nodes, amount, store_identity, tunnel, communityclass, autoload_discovery)
+        return blockingCallFromThread(reactor, _create_nodes, amount, store_identity, tunnel, community_class,
+                                      autoload_discovery, memory_database)
+
+    def _create_node(self, dispersy, community_class, c_master_member):
```
whirm, line 112: Are you using this somewhere else or it's a cleanup leftover?

snorberhuis, line 112: I used it somewhere else.

I used it in my own tests to setup a DebugNode with a specific curve by overriding the method.

tests/dispersytestclass.py
snorberhuis, line 28: This code was normally not executed during tests for dispersy, because a logger.conf is not available by default in Dispersy.
But I started testing from Tribler where one is available by default. This resulted in an exception, because the logging.config is not imported correctly.

If config.fileConfig is used, then during test execution all debug statements are logged. While if basicConfig is used, the tests have clean logging statements.

snorberhuis, line 28: Working for more than a week on this resulted in poor commit etiquette. Sorry!

tests/dispersytestclass.py
whirm, line 112: ok

whirm:
Ready to merge then?

PR #468, opened 2016-04-20
Temporary fork/fix for libnacl

whirm:
Makes it use libsodium18.so if found so it works on recent Debian/Ubuntu versions.


whirm:
retest this please

PR #473, opened 2016-04-26
Include a fallback option for LAN ip discovery when netifaces fails

pimveldhuisen:


dispersy.py
```diff
@@ -795,9 +829,12 @@ def set_connection_type(connection_type):
         if len(self._wan_address_votes[address]) > len(self._wan_address_votes.get(self._wan_address, ())):
             if set_wan_address(address):
                 # refresh our LAN address(es), perhaps we are running on a roaming device
-                self._local_interfaces = list(self._get_interface_addresses())
-                interface = self._guess_lan_address(self._local_interfaces)
-                lan_address = ((interface.address if interface else "0.0.0.0"), self._lan_address[1])
+                if self._netifaces_failed:
+                    lan_address = self._get_lan_address()
+                else:
+                    self._local_interfaces = list(self._get_interface_addresses())
```
whirm: This is only set in one of the branches.

whirm: Also check if they should be a function?

dispersy.py
```diff
@@ -778,7 +807,12 @@ def set_connection_type(connection_type):
 
         # ignore votes from voters that we know are within any of our LAN interfaces.  these voters
         # can not know our WAN address
-        if any(voter.sock_addr[0] in interface for interface in self._local_interfaces):
+        if self._netifaces_failed:
```
whirm: that could be a function

whirm:
Looks good, please squash!

PR #474, opened 2016-04-27
Moved reactor thread check

devos50:
In order to be able to startup Tribler on the (regular) Twisted reactor, `blockingCallFromThread` should check if we are running on the reactor thread already or not. So I moved this check from `blocking_call_on_reactor_thread` to `blockingCallFromThread`.


devos50:
@whirm I will change the dispersy pointer in Tribler on my twisted plugin PR.

whirm:
Ok

PR #477, opened 2016-05-04
Fix 2 bugs in the netifaces fallback

pimveldhuisen:


whirm:
Looks fine.

@pimveldhuisen please remember to set it to READY when you are done with a PR, otherwise it will slip through.

PR #482, opened 2016-05-25
Add an option to read an existing LibNaCL keypair

pimveldhuisen:
There already was an option to generate with the same 'crypt', but to generate the exact same keypairs for both signing and encrypting we also need the same 'seed'.


pimveldhuisen:
Nevermind, I just realized I am a complete idiot here.

PR #490, opened 2016-06-09
Jenkisfile

whirm:


whirm:
@devos50, @lfdversluis, @qstokkink, @pimotte, @pimveldhuisen, @brussee, @synctext

I think it's done.

This Pipeline runs the dispersy tests on linux/osx/windows and the tribler tests on linux with Tribler from `devel` and the PRs dispersy revision to check if it contains changes that break compatibility. If any of the tests fails (except the Tribler ones for now) fails, the test runs will abort immediately so the push->run->fix->push iterations are as fast as possible.

If all tests pass, the job will scan for a free DAS5 cluster and schedule a short run of the AllChannel experiment so we can check if there's any obvious breakage or regression.

It also:
- Generates the global coverage and quality diff reports
- Checks if there's Merge, fixup or DROPME commits on the PR to avoid merging them by accident.

There's 3 flags:

``` groovy
//////////////////
def failFast = true
def skipTests = false
def skipExperiments = false
//////////////////
```

That can be modified by the PR author to temporalily modify the behavior of the pipeline when checking for a particular part to be working properly.

Some other advantages that come with using a Jenkins pipeline:
- Much less clutter on our jenkins instance
- Each PR and branch gets it's own job history which can be destroyed once it gets merged without affecting the others
- Any improvement on the pipeline will go through the PR procedure, so if it breaks, it won't affect the rest of the developers.
- The test runs start quicker as we only clone the repos once and then distribute them to all the involved slaves.

Please, let me know if you feel there's anything missing (have a look at the reports by clicking on the `Jenkins` check).

I plan on merging this and disabling the old dispersy job on Monday/Tuesday next week.

whirm:
Oh, also, Jenkins can be restarted in the middle of a job execution and it will keep its state.

.pylintrc
```diff
@@ -0,0 +1,386 @@
...
+# Specify a configuration file.
+#rcfile=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=.git,dispersy,pymdht,libnacl,data
```
lfdversluis: I think ignoring dispersy on dispersy will leave not much :D

lfdversluis: I think something went wrong with github. In any case: ignoring dispersy on dispersy will not leave much I think :D

tests/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+
+# Code:
+
+adfadsf
```
lfdversluis: debug?

.pylintrc
whirm: Oh, good catch :D

whirm: I copied it from Tribler directly

tests/__init__.py
whirm: That's on the DROPME commit, it's used to make the current PR tester fail right away and not clog the job queue with useless jobs while I fix/improve the pipeline

whirm:
I'm no longer working on this project, @devos50 should eventually finish my work on this. Thanks for the heads up.

whirm:
Closing this, feel free to branch off and finish it

PR #492, opened 2016-06-15
Added sanity check in get_member

lfdversluis:
Watch it burn.


lfdversluis:
FAILED (SKIP=4, errors=396)

lfdversluis:
Might be old unmaintained code. It never gets called in current devel.

whirm:
Thanks for checking. Let's see if I manage to have a look at it on Monday

devos50:
@qstokkink and I found out what was wrong: the check assumes that `my_member` of a community has an associated dispersy-identity message in the database but this is not always true when we put this check inside the `get_member` method.

PR #493, opened 2016-06-15
Added a test that checks if the amount of nodes that are created is as expected

lfdversluis:


lfdversluis:
test_expect_two_nodes (dispersy.tests.test_batch.TestBatch) ... FAIL

devos50:
I found out that an additional member is created when the `create_nodes` is called. This member represents the master member of the community so this is expected behaviour (although this should be documented better).

PR #504, opened 2016-07-10
Commented out assertion

devos50:
I comment this assertion out for now so we can safely increase the Dispersy pointer on Tribler.

### DIFF
--- a/taskmanager.py
+++ b/taskmanager.py
@@ -35,7 +35,9 @@ class TaskManager(object):
         """
         Register a task so it can be canceled at shutdown time or by name.
         """
-        assert isInIOThread()
+        # TODO(Martijn): this assert is necessary, however, we are calling register_task many times from a non-io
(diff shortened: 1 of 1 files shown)


devos50:
@lfdversluis can you review?

lfdversluis:
Do we want to disable this? Who is currently fixing the non-twisted calls? Was it @whirm?

devos50:
@lfdversluis this was suggested by @whirm on IRC. In the end, the assertion should definitely be there but that means we have to wait until all `register_task` calls are fixed before we can adjust the dispersy pointer in Tribler. That's why it's disabled for now.

@whirm is fixing the thread pool stuff, I'm not sure whether he also fixes these calls in the meanwhile.

lfdversluis:
Alright, I understand that as I too was facing this assertion.

whirm:
Yes, I'm fixing the calls too.

devos50:
@whirm nice!

PR #507, opened 2016-08-02
IGNORE: Sql worker test

lfdversluis:
Running a worker on a twisted thread makes backwards compatibility really hard, if not impossible, so I removed it and now test if this works.


lfdversluis:
retest this please

PR #525, opened 2017-02-06
Ensure _get_interface_addresses always returns IP addresses as str

egbertbouman:
On my Windows machine an assert is firing due to an unicode IP address being returned from _get_interface_addresses (instead of a str). Apparently it's due to netifaces returning an unicode object on Windows (https://bitbucket.org/al45tair/netifaces/issues/41/address-info-unicode-on-windows-byte).


egbertbouman:
ok to test

devos50 (approved):

PR #533, opened 2017-04-05
Catch rogue (twisted) threads in unit test teardown

qstokkink:
This PR adds checking for unclosed reactor threads and normal threads on test `tearDown` in `DispersyTestFunc`.


qstokkink:
@devos50 please review if you have time, the failing test is an actual bug which managed to slip through the tests (and should be fixed ASAP of course).

devos50 (approved):

PR #537, opened 2017-04-11
READY: Fix test_tracker unit test

qstokkink:
This PR fixes the `test_tracker` unit test to deal with twisted.
It also auto-inserts the dispersy path in the twistd tracker plugin (which is more convenient than forcing users to do this themselves)*.

*Previously this was done within the `test_tracker` unit test.

### DIFF
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -16,7 +16,7 @@ from ..candidate import Candidate
 from ..dispersy import Dispersy
 from ..endpoint import StandaloneEndpoint
 from ..message import Message, DropMessage
-from ..util import blockingCallFromThread
+from ..util import blockingCallFromThread, blocking_call_on_reactor_thread
 from .debugcommunity.community import DebugCommunity
 from .dispersytestclass import DispersyTestFunc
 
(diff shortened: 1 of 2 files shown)


qstokkink:
@devos50 If you have time, please review/merge.

tests/test_bootstrap.py
```diff
@@ -98,6 +92,9 @@ def dispersy_enable_candidate_walker_responses(self):
                                                                global_time=42),
                               destination)
 
+            # The endpoint is twistified, release the reactor lock for packets to process
```
devos50: I thought the `StandaloneEndpoint` was running on a separate thread outside the reactor?

tests/test_bootstrap.py
```diff
@@ -98,6 +92,9 @@ def dispersy_enable_candidate_walker_responses(self):
...
+            yield deferLater(reactor, 0.1, lambda: None)
```
devos50: Still better than blocking sleeps but still not very pretty :(

devos50 (commented):
Looks ok but I'm wondering about the reason why this deferLater is necessary. Can you elaborate that?

tests/test_bootstrap.py
qstokkink: Good catch, I think this is a legacy fix on my part.

qstokkink:
@devos50 You were correct, the deferLater wasn't necessary.

devos50 (approved):
Nice 👍

PR #538, opened 2017-04-11
NEEDS DISCUSSION: Removed incoming message batching

qstokkink:
This PR removes incoming message batching (it DOES NOT remove the delayed message system).

In Tribler this affects the `tunnel_community` as it uses the `cache` argument for `on_incoming_packets`:
https://github.com/Tribler/tribler/blob/devel/Tribler/community/tunnel/tunnel_community.py#L1041

As shown by the unit tests, there are cases where this approach causes a large increase in the amount of messages being sent.


triblerteam:
No merging? Feel free to remove message batching at some point using some approach; assuming it does no harm.

qstokkink:
@triblerteam This -and other wild optimizations- will be part of [IPv8](https://github.com/qstokkink/py-ipv8).

PR #544, opened 2017-06-03
WIP: Remove M2Crypto

vandenheuvel:
- [x] Removing M2Crypto
- [ ] Minor issues in the code
- [ ] Updating CI


devos50:
ok to test

devos50:
Note that we might want to use lower-security curves during experiments so I would at least keep the possibility to use curves that are faster. @qstokkink what is your opinion on this?

qstokkink:
@devos50 We use that a lot, so I would say we keep that feature.

vandenheuvel:
retest this please

vandenheuvel:
retest this please

qstokkink:
Here's how I did this in `ipv8`:

https://github.com/qstokkink/py-ipv8/blob/master/keyvault/public/m2crypto.py
https://github.com/qstokkink/py-ipv8/blob/master/keyvault/private/m2crypto.py

PR #549, opened 2017-06-08
READY: Removed sleeps from tests

qstokkink:
This PR twistifies all sleep calls in the unit test code, EXCEPT in `test_bootstrap` (as this uses sleeps to control external processes). 

I recommend reviewing `test_double_signature` in [split diff mode](https://github.com/Tribler/dispersy/pull/549/files?diff=split), as this class had all of its Widows line endings (\r\n) replaced with Nix endings (\n).

NOTE: The multichain Tribler test will have to be updated to work with this PR.

### DIFF
--- a/endpoint.py
+++ b/endpoint.py
@@ -378,6 +378,11 @@ class ManualEnpoint(StandaloneEndpoint):
         self.process_packets(packets)
         return packets
 
-    def process_packets(self, packets, cache=True):
+    def process_packets(self, packets, cache=False):
         self._logger.debug('processing %d packets', len(packets))
-        StandaloneEndpoint.data_came_in(self, packets, cache=cache)
+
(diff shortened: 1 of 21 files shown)


qstokkink:
@devos50 Hmm.. it seems that this twistified PR:
https://jenkins.tribler.org/job/GH_Dispersy_PR_tests_linux/37/testReport/
is actually **SLOWER** across the board than the `sleep()` tests:
https://jenkins.tribler.org/job/Test_dispersy_devel/lastCompletedBuild/testReport/

Do we really want this?

devos50:
Interesting, I didn't expect that!

At least the Dispersy tests are twistified now but if they becomes slower, we have to wait even longer.

qstokkink:
@devos50 We may just have to swallow our pride on this one and keep the sleeps in.

PR #556, opened 2017-06-23
READY: Rename ManualEnpoint to ManualEndpoint

vandenheuvel:
There is a typo in the name of the `ManualEnpoint` class.


devos50:
ok to test

devos50:
@vandenheuvel have you checked that this `ManualEndpoint` is used somewhere in Tribler and/or Gumby?

vandenheuvel:
@devos50 It is being used in Tribler, I haven't yet looked at Gumby. However, the dependent projects can be updated when the Dispersy pointer is increased.

PR #574, opened 2018-12-02
Disabled bloomfilter check

devos50:
See https://github.com/Tribler/tribler/issues/2726


qstokkink (approved):

devos50:
Unit tests passed but there was an error when parsing the `nosetests.xml` file. I've fixed it 👍

